PR #299, opened 2015-01-11
Fix include flags for LibGcrypt

ysangkok:
When it's not installed in a default path, the compiler calls will not find Gcrypt unless we add it like this.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -158,6 +158,7 @@ endif()
 find_package(GCRYPT REQUIRED)
 add_definitions(-DHAVE_GCRYPT)
 list(APPEND EXTRA_LIBS ${GCRYPT_LDFLAGS})
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${GCRYPT_CFLAGS}")
 
 # Check for GnuTLS
 if(ENABLE_GNUTLS)


CMakeLists.txt
```diff
@@ -158,6 +158,7 @@ endif()
 find_package(GCRYPT REQUIRED)
 add_definitions(-DHAVE_GCRYPT)
 list(APPEND EXTRA_LIBS ${GCRYPT_LDFLAGS})
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${GCRYPT_CFLAGS}")
```
flashcode: I think it's better (or more "clean") to add definitions instead of changing directly CFLAGS, can you try that: `add_definitions(${GCRYPT_CFLAGS})`.
I think the add_definitions should work to compile tests too (which are written in C++).
If that work fine, please rebase your pull request with that instead.

flashcode:
Some news about my suggestion?
And another question: on which system or which case you got the libgcrypt compilation problem? If you specify a custom location for the lib?

ysangkok:
It seems to work, but the rpath isn't set correctly, and it can't find `libgcrypt.so.20` on startup time.

Here's the cmake line I'm using:

    ~/cmake-3.1.0-Linux-x86_64/bin/cmake . -DGCRYPT_CFLAGS="`~/libgcrypt-build/bin/libgcrypt-config --cflags`" -DGCRYPT_LDFLAGS="`~/libgcrypt-build/bin/libgcrypt-config --libs`" -DCURL_LIBRARY="~/curl-build/lib/libcurl.so" -DCURL_INCLUDE_DIR="~/curl-build/include" -DCMAKE_INSTALL_PREFIX:PATH=~/weechat-build

The system is Ubuntu, but I don't have admin rights, so I install all packages in my home directory. By the way, I think the other dependencies have issues too, but I really only need a minimal build.

Should I rebase with your suggested patch or do you want to fix the rpath issue too? I don't know how to do that, since we are already handling the LDFLAGS correctly, it seems.

flashcode:
I'll have to investigate.

But why do you need a custom build of gcrypt? This sounds like a not so common problem, especially on Debian/Ubuntu where it's easy to install binary packages of weechat (stable or devel version).

ysangkok:
I don't need a custom version, but since I can't install the deb packages without root, I need to build my own gcrypt since the development headers aren't installed. Also, custom library locations are useful when cross-compiling.

flashcode:
Maybe the RPATH problem is related to this bug: https://savannah.nongnu.org/bugs/?41525 ?

flashcode:
Not sure if that's still needed, closing the issue.

PR #300, opened 2015-01-12
doc: mention otr.py in English FAQ (REF: #298)

Mikaela:


Mikaela:
I updated this pull request to reflect `requirements.txt` which was added to weechat-otr in case that there will be more requirements in the future.

I specify the flags `--user --upgrade` to not have issues with possibly missing root access and in case there are older versions installed (it won't do anything if there is same version installed) and I am used to typing it.

Also, on \#weechat-otr:

```
2015-01-13 22:41:20+0200 < tribut> not really sure if the faq is the best place to explain the debian package thing. maybe just a link like "for installation instructions, please visit the plugin's website" and then link to our readme?
```

flashcode:
The purpose of FAQ is not to mention all possible scripts, otherwise it would become bloated.

The quickstart and user's guide give instructions to browse the scripts repository, which should be enough to find what you want, for example OTR support.

So I'm closing this issue.

Mikaela:
There is always https://github.com/weechat/weechat/issues/298

flashcode:
Yes OTR as native plugin would be great.

PR #301, opened 2015-01-12
doc: FAQ: use all options with homebrew

Mikaela:
It can be confusing for OS X users to have different packages installed
than what ship with most of WeeChat packages on other operating systems.

I added `brew install aspell` on top, because of the warning in
`brew info weechat` saying that `If Aspell was installed
automatically as part of weechat, there won't be any dictionaries.`

### DIFF
--- a/doc/en/weechat_faq.en.asciidoc
+++ b/doc/en/weechat_faq.en.asciidoc
@@ -63,7 +63,8 @@ brew info weechat
 You can install WeeChat with this command:
 
 ----
-brew install weechat --with-python --with-perl
+brew install aspell
+brew install weechat --with-aspell --with-curl --with-guile --with-lua --with-perl --with-python --with-ruby --with-tcl
 ----
 
 [[lost]]


doc/en/weechat_faq.en.asciidoc
```diff
@@ -63,7 +63,8 @@ brew info weechat
 You can install WeeChat with this command:
 
 ----
-brew install weechat --with-python --with-perl
+brew install aspell
+brew install weechat --with-aspell --with-curl --with-guile --with-lua --with-perl --with-python --with-ruby
```
flashcode: Isn't there a `--with-tcl` flag in homebrew for WeeChat?

Mikaela: From linuxbrew (I can check OS X when someone boots it, but both brews should be the same):

```
% brew info weechat
weechat: stable 1.1, HEAD
https://www.weechat.org
Not installed
From: https://github.com//homebrew/blob/master/Library/Formula/weechat.rb
==> Dependencies
Build: cmake ✘
Required: gnutls ✘, libgcrypt ✘, gettext ✘
Optional: guile ✘, aspell ✘, lua ✘, curl ✘
==> Options
--with-aspell
        Build with aspell support
--with-curl
        Build with brewed curl
--with-guile
        Build with guile support
--with-lua
        Build with lua support
--with-perl
        Build the perl module
--with-python
        Build with python support
--with-ruby
        Build the ruby module
--HEAD
        Install HEAD version
==> Caveats
Weechat can depend on Aspell if you choose the --with-aspell option, but
Aspell should be installed manually before installing Weechat so that
you can choose the dictionaries you want.  If Aspell was installed
automatically as part of weechat, there won't be any dictionaries.
```

Mikaela: Actually the homebrew formula doesn't mention tcl either. https://github.com/Homebrew/homebrew/blob/master/Library/Formula/weechat.rb

flashcode:
Then maybe a `--with-tcl` should be added?
By the way, gnutls is optional in WeeChat, so it should not be forced (it could be compiled by default, but you should be able to disable it, since WeeChat compiles fine without).

Mikaela:
> Then maybe a --with-tcl should be added?
> By the way, gnutls is optional in WeeChat, so it should not be forced (it could be compiled by default, but you should be able to disable it, since WeeChat compiles fine without).

These both should be reported at https://github.com/homebrew/homebrew/issues or send a pull request to that file I linked earlier.

I think that if homebrew doesn't automatically require gnutls, most of homebrew users will be missing SSL support and come to IRC asking about that. TCL seems like an issue to me.

flashcode:
I opened a pull request: https://github.com/Homebrew/homebrew/pull/36186

Mikaela:
Added assuming that your pull request gets merged.

flashcode:
Merged without the `--with-tcl` since it seems it's not needed (pull request was closed).

tdsmith:
> I added brew install aspell on top, because of the warning in
brew info weechat saying that If Aspell was installed automatically as part of weechat, there won't be any dictionaries.

FWIW, that's an error in the weechat caveats that I should fix; you'll always get (US) English. `brew install weechat --with-aspell` gives you the same aspell as `brew install aspell` if you haven't already had aspell installed. The moral of the message is really just that if you need e.g. Finnish you should `brew install aspell --with-lang-fi` first.

PR #325, opened 2015-01-31
relay: update relay.network.allowed_ips desc.

anders:
v4-mapped addresses aren't used since commit 85339f6.


flashcode:
Merged.

PR #329, opened 2015-02-03
Update irc-nick.c

zvo:
weechat breaks on networks that allow color codes in nicknames, this fixes that.
see: http://savannah.nongnu.org/bugs/?41416

### DIFF
--- a/src/plugins/irc/irc-nick.c
+++ b/src/plugins/irc/irc-nick.c
@@ -84,8 +84,6 @@ irc_nick_is_nick (const char *string)
 
     while (ptr && ptr[0])
     {
-        if (!strchr (IRC_NICK_VALID_CHARS, *ptr))
-            return 0;
         ptr++;
     }
 


flashcode:
Hi,
Could you please tell me how to reproduce the problem?
The irc server used and the steps to see the problem.
Thanks.

zvo:
Hey,
1. Connect to a network/server that allows color codes in nicknames (e.g irc.gangnet.org). 
2. Try changing your nick to something with colors using /nick 
3. Change your nick again to something else.
4. Watch everything break (joining channels, messaging, etc)

There's also other issues such as tabcomplete not coping with 0x03 

Here's a source for the IRC server used on the network above: https://github.com/malevolm/ghorme-sabzircd  . There exists several other ones with similar functionality.

flashcode:
After looking at the code, the function you updated (irc_nick_is_nick) is called only in info "irc_is_nick", which can be called by scripts only.
So your fix could help with scripts, but I'm not sure it can help with WeeChat itself.

flashcode:
@zvo: some news about my comment above?
As I said, I don't understand how your commit could fix the problem you mentioned.

flashcode:
Closing the pull request, for me the fix is not good, and no news since Feb 13.

PR #342, opened 2015-02-28
Fix some strings for pt_BR.po using msgcheck

camponez:
This is an update to the pt_BR language file


flashcode:
Hi Eduardo,
Thanks for the update on these translations.
I see in the diff that you translated/fixed some translations which are still marked as fuzzy.
Is that normal? Strings fixed should not be "fuzzy" anymore.

camponez:
@flashcode I kind of let them fuzzy on purpose. But know I think all the strings has passed the check.

Issue #187, opened 2014-09-02
Ignore certain ("non-nick") characters in nick completion


scajanus:
Currently the nick completion assumes that the nick begins with either beginning of input or space. In some situations, it would make sense to ignore other characters as well, e.g. in case of addressing a person inside parentheses.

Example:

> Text addressed to channel in general, but with personal comment to a specific nick. (nick_2: that would make sense.)

In this case the nick completion does not work. It would make sense to have this option as a list of chars that would be ignored, unless there's a nick that has those characters, in which case you it might make sense first to complete that one, and after that the other options.

(An option named weechat.completion.nick_ignore_chars exists, but it has different functionality, where it completes nick to _nick1 if _ is included.)

smaslennikov:
This is quite needed for slack gateways; people don't get pinged without being mentioned with their nick prepended by '@'.

PR #345, opened 2015-03-05
core: add config option to ignore prefix characters in completions

irisfofs:
This configuation option (config_completion_ignore_prefix_chars) makes the specified characters (default: "@():") be treated as though they were spaces when performing completions. So "@Ni" completes to "@Nick", etc.

affected functions:
- gui_completion_find_context

### DIFF
--- a/src/core/wee-config.c
+++ b/src/core/wee-config.c
@@ -246,6 +246,7 @@ struct t_config_option *config_color_nicklist_offline;
 struct t_config_option *config_completion_base_word_until_cursor;
 struct t_config_option *config_completion_command_inline;
 struct t_config_option *config_completion_default_template;
+struct t_config_option *config_completion_ignore_prefix_chars;
 struct t_config_option *config_completion_nick_add_space;
 struct t_config_option *config_completion_nick_completer;
 struct t_config_option *config_completion_nick_first_only;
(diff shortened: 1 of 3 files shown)


flashcode:
I'm OK for such new option, but I think it should be restricted to nick completions.
Because with your code, if you do `/bar @a` +Tab key, this will complete to `/bar @add` which is not valid.
I think it's only useful in nick completions. And if ever we need that elsewhere, then we should think about more options to control these prefix chars.

So if you're OK I can merge this pull request and then restrict the option to nick completions.

xiagu:
I think you're right. I created it for use with nick completions after all (@mentions in BitlBee, primarily).

Merging and restricting it to just nick completions would be fine.

jdve:
This is a great addition! :+1:

skyhighwings:
:+1:

flashcode:
This pull request (if merged) will close issue #187.

deesee04:
any idea when this might be merged and released? 

besides the bitlbee stuff, i use slack extensively for work and thus this is my killer feature!

xiagu:
@deesee04 Depending how much of a killer feature it really is, you can always build the fork from source: https://github.com/xiagu/weechat

I suspect that needing to edit it after merging is one thing holding it back. When I have some free time, I'll look into restricting it to just nick completions myself. (And maybe address a point brought up in #187 -- when nicks contain an ignored character. I've never seen a network allow any of `@():` in nicks, but ideally I think you'd want the union of nick completions with and without the ignored characters.)

deesee04:
@xiagu totally, although i try to stick to the mainline releases. this issue is not a showstopper, just incredibly frustrating :)

i really appreciate the work on this item!

flashcode:
This feature is planned for version 1.3.

deesee04:
as this isn't due to hit the mainline branch until august, i went ahead and built this fork. works well - a true lifesaver for some use cases!

one note - tab completing with an ignored character adds just a whitespace at the end. it would be nice if it could still end with a colon and whitespace, as the official tab completion mechanic does.

flashcode:
@deesee04: even if the stable version if planned for August, you can use the development version, it is (almost always) very stable, and you can find nightly builds on weechat.org (sources + debian/ubuntu/raspbian packages).

xiagu:
I'm working on this again, but progress is in the [feature/ignore-prefix-characters](/xiagu/weechat/tree/feature/ignore-prefix-characters) branch of my fork so as not to spam this with updates. Later I'll replace this branch with a squashed version and all that good stuff.

PR #356, opened 2015-03-09
script: use https for script.scripts.url

### LINKED ISSUE #355: Make script.scripts.url = "https://www.weechat.org/files/plugins.xml.gz" as default.
As of now the default setting is 
`script.scripts.url = "http://www.weechat.org/files/plugins.xml.gz"`

Wouldn't it be safer to have weechat download the scripts (library) using HTTPS instead of HTTP?

Mikaela:
Closes #355

### DIFF
--- a/src/plugins/script/script-config.c
+++ b/src/plugins/script/script-config.c
@@ -704,7 +704,7 @@ script_config_init ()
         script_config_file, ptr_section,
         "url", "string",
         N_("URL for file with list of scripts"),
-        NULL, 0, 0, "http://weechat.org/files/plugins.xml.gz", NULL, 0, NULL, NULL,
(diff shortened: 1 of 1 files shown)


Mikaela:
2015-03-09 17:34:14+0200 <@FlashCode> Mikaela: weechat already uses https by default for scripts, with new option: /set script.scripts.url_force_https
2015-03-09 17:34:24+0200 <@FlashCode> this overrides http:// to https:// automatically
2015-03-09 17:34:27+0200 < Mikaela> what is that option then?
2015-03-09 17:34:28+0200 <@FlashCode> when downloading
2015-03-09 17:34:29+0200 < Mikaela> oh
2015-03-09 17:34:38+0200 <@FlashCode> so you can close both PR and issue I think :)
2015-03-09 17:34:46+0200 < Mikaela> the issue is not by me
2015-03-09 17:35:11+0200 < Mikaela> maybe that option should be mentioned in help text

PR #367, opened 2015-03-22
trigger: hide password when messaging NickServ through default alias

sim642:
There is a default alias `M => msg` which is easy to type when interacting with NickServ for example but the passwords were only hidden when using the non-alias. This pull request fixes that.


flashcode:
Thanks, I forgot this default alias for `/msg`.

PR #370, opened 2015-03-23
Update mail in copyright comments.

KokaKiwi:
I don't know if it's really useful, but as the old mail doesn't exist anymore, I think the new mail address should figured in place of the old one.


KokaKiwi:
Also, maybe I shouldn't specify my name in the commit message... :x

flashcode:
Merged.

Issue #385, opened 2015-04-07
RFC: enhancement: native RobustIRC support


stapelberg:
I’m one of the developers of http://robustirc.net/. See http://robustirc.net/docs.html for more details, but in a nutshell, it’s an IRC network written on top of https://raftconsensus.github.io/ which doesn’t suffer from netsplits. To not be affected by network issues between the client and the IRC servers, clients should use the RobustSession protocol which wraps IRC in a connection-independent HTTPS/JSON-based transport, see http://robustirc.net/docs/robustsession.html for details.

Currently, we offer a program called the RobustIRC bridge, which you can configure as a proxy server in WeeChat. The bridge then does the conversion (bridging) from IRC to IRC in RobustSession.

I intend to submit a pull request for WeeChat to talk RobustSession natively, i.e. without the bridge. Would such a pull request be accepted?

Thank you.

flashcode:
Hi Michael,
First thanks for i3, I love your window manager ;)
I didn't read everything about RobustSession yet, but if I understand correctly it would be an encapsulation of IRC message in another protocol, so it should impact almost only the way to send/receive messages in irc plugin (network stuff), am I right?
Feel free to send a pull request, and we'll discuss there about the changes.

Shawn-Smith:
I skimmed through the docs a bit. It's quite a big change and I think it'd be better off as a new plugin rather than added to the irc plugin.

stapelberg:
> First thanks for i3, I love your window manager ;)

Glad to hear you like it :).

> I didn't read everything about RobustSession yet, but if I understand correctly it would be an encapsulation of IRC message in another protocol, so it should impact almost only the way to send/receive messages in irc plugin (network stuff), am I right?

You’re correct — it’s only a transport layer, and the client still speaks IRC over it just as it does today.

> I skimmed through the docs a bit. It's quite a big change and I think it'd be better off as a new plugin rather than added to the irc plugin.

We briefly evaluated this before (by just reading the WeeChat plugin docs) and didn’t see an easy way to say that our protocol should be based on the IRC plugin, just with a different transport. Can you outline how we’d go about implementing this in a separate plugin please?

Mikaela:
Is RobustIRC official standard?

The next version of IRC is [IRCv3](http://ircv3.org) and it's development is happening at \#ircv3 on chat.freenode.net. You might want to contact them.

kaniini:
[...]

PR #387, opened 2015-04-10
WIP: robustirc support

stapelberg:
This is a very very rough first proof of concept implementation of how
native RobustIRC support in WeeChat could work.

The intention of this commit is to serve as a base for discussion, not
to be used or even committed to the repository.

### DIFF
--- a/src/plugins/irc/irc-server.c
+++ b/src/plugins/irc/irc-server.c
@@ -61,6 +61,7 @@
 #include "irc-notify.h"
 #include "irc-protocol.h"
 #include "irc-raw.h"
+#include "irc-robustirc.h"
 #include "irc-redirect.h"
 #include "irc-sasl.h"
 
@@ -113,6 +114,7 @@ char *irc_server_options[IRC_SERVER_NUM_OPTIONS][2] =
   { "default_msg_part",     "WeeChat %v"          },
   { "default_msg_quit",     "WeeChat %v"          },
   { "notify",               ""                    },
+  { "robustirc",            ""                    },
 };
 
 char *irc_server_casemapping_string[IRC_SERVER_NUM_CASEMAPPING] =
@@ -1050,6 +1052,8 @@ irc_server_alloc (const char *name)
     new_server->buffer_as_string = NULL;
     new_server->channels = NULL;
     new_server->last_channel = NULL;
+    new_server->robustirc_sessionid = NULL;
+    new_server->robustirc_sessionauth = NULL;
 
     /* create options with null value */
     for (i = 0; i < IRC_SERVER_NUM_OPTIONS; i++)
@@ -1853,6 +1857,9 @@ irc_server_send (struct t_irc_server *server, const char *buffer, int size_buf)
         return 0;
     }
 
+    if (IRC_SERVER_OPTION_BOOLEAN(server, IRC_SERVER_OPTION_ROBUSTIRC))
+        return irc_robustirc_send(server, buffer, size_buf);
+
 #ifdef HAVE_GNUTLS
     if (server->ssl_connected)
         rc = gnutls_record_send (server->gnutls_sess, buffer, size_buf);
@@ -2504,7 +2511,7 @@ irc_server_msgq_flush ()
         if (irc_recv_msgq->data)
         {
             /* read message only if connection was not lost */
-            if (irc_recv_msgq->server->sock != -1)
+            if (irc_recv_msgq->server->sock != -1 || IRC_SERVER_OPTION_BOOLEAN(irc_recv_msgq->server, IRC_SERVER_OPTION_ROBUSTIRC))
             {
                 ptr_data = irc_recv_msgq->data;
                 while (ptr_data[0] == ' ')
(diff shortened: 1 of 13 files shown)


src/plugins/irc/irc-server.c
```diff
@@ -3387,7 +3439,7 @@ irc_server_connect_cb (void *data, int status, int gnutls_rc, int sock,
             server->current_ip = (ip_address) ? strdup (ip_address) : NULL;
             weechat_printf (
                 server->buffer,
-                _("%s%s: connected to %s/%d (%s)"),
+                "%s%s: connected to %s/%d (%s) lolwat",
```
Shawn-Smith: You've added this "lolwat" to the code in two spots, what is it for?

stapelberg: That was from the early stages of debugging, when I was still trying to get my change/compile/test workflow working for WeeChat. Please ignore :).

stapelberg:
@flashcode So, what do you think about this pull request?

I’ve also tried coming up with an out-of-tree version of a RobustIRC plugin for WeeChat. The best approach I could find without modifying the WeeChat code is to add a signal hook for `irc_server_connecting`, then use `irc_server_search` and unhook `server->hook_connect` to abort the current connection attempt. I imagine (but haven’t tried) that I can call `irc_server_connect_cb()` with a file descriptor that I have opened in my plugin (e.g. a pipe).

The problem with that approach is that it depends on the irc plugin internals which are not exposed in the plugin API currently.

How do you think we could best extend the plugin API to make a custom IRC transport plugin possible?

(FWIW, in the meantime I have a working version of an out-of-tree irssi module, see https://github.com/robustirc/irssi-robustirc — I would love to also have one for WeeChat)

flashcode:
Hi @stapelberg,

Sorry for the long time to reply, I was focused on other bugs and new features in WeeChat for the release of v1.2 (scheduled in two days).

So I've read RobustIRC protocol, discussions in the issue, and this pull request.

First some comments on this pull request, even if your intention is not to commit it as-is:

* The WeeChat "hooks" are protocol agnostic and low-level, so for me `hook_connect_robustirc` should not be there (there's no IRC protocol nor any other protocol in WeeChat core); `hook_fd` should be enough for socket communication, and then a plugin can implement layers over that.
* You use Curl in RobustIRC, this may block WeeChat sometimes, even if you use the multi interface of curl (see: http://curl.haxx.se/libcurl/c/libcurl-multi.html#BLOCKING); currently all network operations that can block are implemented with a fork in `wee-network.c`.
* The WeeChat plugins must never include or rely on code of WeeChat core; the only file that can be used by plugins is `weechat-plugin.h`; in other words that means any WeeChat plugin can be compiled without even having the sources of WeeChat (only `weechat-plugin.h` is required).
* Minor thing, I know you were not focused on that: automatic tests are broken by this pull request, because WeeChat can still be compiled with autotools (for historical reasons, but now CMake is preferred).

Now about the fact to include RobustIRC inside IRC plugin, as a separate plugin/script, or even as a separate proxy... this is the harder part, because I've not the best solution in mind yet.

Here are the pros and cons for me of each solution (the list is not exhaustive and we can discuss about each point):

#### Add RobustIRC inside IRC plugin (this pull request)

##### Pros

* Good integration with IRC plugin, just a boolean to toggle, and then you can use RobustIRC.

##### Cons

* As said in issue comments, the IRC plugin may become "bloated", with the add of Curl, HTTP, JSON and such in IRC plugin, which is very different from the original IRC protocol.
* It will not be possible to use RobustIRC with other protocols than IRC in WeeChat (anyway, can RobustIRC support other protocols? or it should be renamed :wink:).
* It will be harder to remove this feature (if we don't need it any more in future), because the code will be mixed with IRC plugin.
* There seem to have plans to integrate more or less similar solutions in future versions of IRC, so maybe we can wait a bit before integrating RobustIRC? (see #385)

#### Add RobustIRC as a separate C plugin

##### Pros

* Better separation of code, and if you don't need the feature, you can just not load the plugin.

##### Cons

* There's currently no way to add or replace a "transport" layer in WeeChat (especially when defined by another plugin), but we can discuss to add that: the way to do it, and to interact with other plugins (IRC in our case).
* Harder to fully integrate in IRC plugin as a server option: not easy to just add a boolean option to the server if the transport layer is managed by an external plugin.

#### Add RobustIRC as a script

##### Pros

* Faster to implement (if the API allows changes in transport layer, see the C plugin above).

##### Cons

* Too slow for me: running script callbacks use a lot of CPU, and this will be called very often, so not an acceptable solution for me.

#### Use RobustIRC as a proxy (external software, not in WeeChat)

##### Pros

* No changes in WeeChat (pros or cons? you may not agree with me here :smile:).
* Easy to setup, with the `.proxy` option in IRC server (and it can be added as a default proxy too with option `irc.server_default.proxy`, so no need to add it in all servers).
* Transparent for the client, this solution works with any IRC client (that at least supports proxy connection).

##### Cons

* Requires the external proxy to run somewhere (local or remote host).
* Adds an extra network layer between WeeChat and the RobustIRC server (if I understood well).
* It would be harder to use another proxy, like HTTP or SOCKS (not sure if the "RobustIRC bridge" itself supports proxy connection to the IRC servers).

--

In short:

* **IRC plugin**: very different from IRC protocol, joker for now on this solution.
* **Separate C plugin**: requires some work in API.
* **Script**: too slow (and requires some work in API).
* **Proxy**: already working, so we can live with that for now.

As a conclusion, I wonder if the proxy solution is not the best, are there really drawbacks to use RobustIRC as a proxy? It would be similar to services like Bitlbee, that you can install separately and connect on from the IRC client (even if RobustIRC is a bit low level compared to Bitlbee).

@stapelberg: it would be nice if we could discuss about all these points, or other things, on IRC (#weechat channel on freenode). I find that more convenient than this issue, especially for long discussions.

JohnMaguire:
I know this ticket is five years old, but I wanted to respond to this bit:

> As a conclusion, I wonder if the proxy solution is not the best, are there really drawbacks to use RobustIRC as a proxy? It would be similar to services like Bitlbee, that you can install separately and connect on from the IRC client (even if RobustIRC is a bit low level compared to Bitlbee).

As an IRC user, the downside to this is that I have to learn and install separate software, and somehow daemonize it, which makes it more difficult to adopt. Integration in a Weechat plugin would allow me to try this out far more easily.

PR #391, opened 2015-04-11
Respect environment variable WEECHAT_HOME

hashken:
If environment variable WEECHAT_HOME is set to a non-empty value, this
is taken as the path for Weechat home directory.

"--dir" command-line switch has higher precedence over this
environment variable.

### DIFF
--- a/src/core/weechat.c
+++ b/src/core/weechat.c
@@ -270,6 +270,53 @@ weechat_parse_args (int argc, char *argv[])
 }
 
 /*
+ * Helper function for weechat_create_home_dir.
+ * Expands and assigns given string to weechat_home
+ */
+
+void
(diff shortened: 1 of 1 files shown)


hashken:
This is especially useful for people who want to keep their home directory in accordance to XDG conventions.

hashken:
Will this be merged? Do you need anymore details?

flashcode:
I'll test that very soon and come back to you if ever I have problems.
According to what I see, this is going to be merged (no problem for me).

viviskas:
> If environment variable WEECHAT_HOME is set to a non-empty value, this
> is taken as the path for Weechat home directory.
> 
> "--dir" command-line switch has higher precedence over this
> environment variable.
>

PR #393, opened 2015-04-11
core: scroll position obeying buffer searching (closes #76)

### LINKED ISSUE #76: Search start position + stop search at point
If I start a search with `Ctrl`+`r`, it scrolls me to the bottom of the buffer, from where I search. This is not necessarily desirable.

If I stop a search with `Enter` (equivalent to  or `Ctrl`+`m`) or `Ctrl`+`j`, it likewise scrolls me to the bottom of the window. There should be a way to stop a search on the point I was looking for.

Consider this redesign of the search interface:
* `Ctrl`+`r` starts a search from anywhere;
* `Enter` stops a search at a certain point (where the cursor is), like in GNU Emacs
* `Ctrl`+`j` returns to the point from where the search began
 * You could also make a new key to return to the bottom, for the convenience of users who like the current behaviour, but I don't see much of a use case for it given how easy it is to get to the bottom by conventional means.

sim642:
This is my attempt on making buffer searching more intuitive by fixing the problems of #76.

Adds `search_text_here` and `search_stop_here` actions for `/input` to not change old behavior but makes them the new defaults. Adds ctrl-Q for old search stop behavior.


src/gui/gui-buffer.h
```diff
@@ -178,6 +178,7 @@ struct t_gui_buffer
     regex_t *text_search_regex_compiled; /* regex used to search            */
     int text_search_where;             /* search where? prefix and/or msg   */
     int text_search_found;             /* 1 if text found, otherwise 0      */
+    struct t_gui_line *text_search_start_line; /* starting line for search  */
```
flashcode: This variable must be moved in struct t_gui_window_scroll (in gui-window.h), because you can perform different search in the same buffer, if it is displayed in multiple windows at same time; that's why it must be stored in the window (with other scroll variables).
And if possible, please rebase all your commits into a single one, thanks!

src/gui/gui-buffer.c
```diff
@@ -4421,6 +4427,7 @@ gui_buffer_print_log ()
         log_printf ("  text_search_regex_compiled: 0x%lx", ptr_buffer->text_search_regex_compiled);
         log_printf ("  text_search_where . . . : %d",    ptr_buffer->text_search_where);
         log_printf ("  text_search_found . . . : %d",    ptr_buffer->text_search_found);
+        log_printf ("  text_search_start_line: 0x%lx", ptr_buffer->text_search_start_line);
```
flashcode: Please align the value with other variables above and below (anyway, this variable will move to gui-window.c).

KamranMackey:
@sim642 yo, you have a few merge conflicts. you should fix them. :D

sim642:
@KamranMackey The conflict has appeared well after my PR and it's in the `hook_command` call of `input` in a width-formatted list of possible actions. Due to the nature of that section of code, there's no nice way for git to handle it and no point in me trying to keep my PR mergeable as any change in there would completely conflict again. It's better fixed at the point of final merge.

PR #410, opened 2015-04-22
no longer put weechat version at the top of configs

### LINKED ISSUE #407: Remove WeeChat version from config files
WeeChat currently puts it's current version at the top of configuration files. This however makes it unpleasant to keep your configuration files in a repository. WeeChat appears to not use this version for anything, and I don't know any usecases myself that would result in needing this.
Can you either remove the functionality that puts this version number in the WeeChat configs, or make it an option to disable it?

Tyil:
I had some spare time at work, so I tried to find the piece of code that actually puts the weechat version into the config and removed it. This is to fix #407

### DIFF
--- a/src/core/wee-config-file.c
+++ b/src/core/wee-config-file.c
@@ -2229,16 +2229,6 @@ config_file_write_internal (struct t_config_file *config_file,
         goto error;
     }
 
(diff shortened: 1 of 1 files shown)


flashcode:
Thanks, fixed by another commit (I kept the comment on top, without the version).

PR #411, opened 2015-04-24
core: add option weechat.look.prefix_unalign_multiline_words

celeron55:
This solves the multi-line URL copying issue that has kept me from migrating to WeeChat from Irssi. It brings infinite and glorious convenience to every user who isn't a fan of vertical splits!

The point of this is that you can use weechat.look.align_end_of_lines = message to get nice formatting, but words that wrap to multiple lines are formatted like weechat.look.align_end_of_lines = time. This allows long URLs to be clicked and the full word to be conveniently copy-pasted.

You may still need weechat.look.eat_newline_glitch = on, depending on how and which terminal you use.

Works with both, weechat.look.prefix_align = none and right. Most of the code is about coping with the slight complexities of the latter.

Works with vertical splits too, with limited usefulness.

For reference, this explains the previously available options for viewing long words as continuous text (that are not sufficient for me and some other users): https://weechat.org/files/doc/weechat_faq.en.html#urls

### DIFF
--- a/src/core/wee-config.c
+++ b/src/core/wee-config.c
@@ -161,6 +161,7 @@ struct t_config_option *config_look_prefix_buffer_align_more;
 struct t_config_option *config_look_prefix_buffer_align_more_after;
 struct t_config_option *config_look_prefix_same_nick;
 struct t_config_option *config_look_prefix_suffix;
+struct t_config_option *config_look_prefix_unalign_multiline_words;
 struct t_config_option *config_look_quote_nick_prefix;
 struct t_config_option *config_look_quote_nick_suffix;
 struct t_config_option *config_look_quote_time_format;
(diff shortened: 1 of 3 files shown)


KamranMackey:
@flashcode you should merge this in. Seems pretty nice. And it's always good to have more customization. :smiley:

mvirkkunen:
Yes @flashcode, you should definitely merge this. This would finally be one actually usable solution to the URL problem.

torhve:
I have tested this patch and it seems to work as advertised if the URL linked is the first word. However if the URL appears at the end of a sentence then I get prefix lines and thus are unable to click it.

torhve:
A screenshot to illustrate what I mean:

Mikaela:
Have you tried `alt + l` (if it does nothing `/key missing`) or `/window bare`? I cannot remember encountering links that didn't work with it.

mvirkkunen:
@Mikaela I'm sure everybody is aware of that option. It's inconvenient compared to being able to just click on links.

celeron55:
Hmm, that seems strange, I must have tested that case too. I'll have to find time to look into this more.

torhve:
@celeron55 there might be a combination of options that's required to make it work? I might have some settings in my client that overrides yours? Can you suggest settings to make it work as intended as the screenshot above displays?

Marqin:
Has anyone looked at this patch and bug that @torhve mentioned?

flashcode:
#802 has been merged.

lidel:
This is a neat feature, thank you!

My `urxvt` was not detecting the end of URL (that was in new line) when running inside of `tmux`,
but suggested `/set weechat.look.eat_newline_glitch on` fixed it and all works perfect now.

flashcode:
@lidel: enabling option `weechat.look.eat_newline_glitch` is not recommended, it can cause severe display bugs.
If not for you, then you're lucky and just forget this comment ;-)

PR #431, opened 2015-05-31
Add default for Elemental-IRCd +y to be Unreal +q

Xe:
Elemental-IRCd has mode +q in use as channel list-like quiet mode. As such it uses +y for owner. This change updates the defaults to reflect this.

### DIFF
--- a/src/plugins/irc/irc-config.c
+++ b/src/plugins/irc/irc-config.c
@@ -2891,7 +2891,7 @@ irc_config_init ()
            "found, WeeChat will try with next modes received from server "
            "(\"PREFIX\"); a special mode \"*\" can be used as default color "
            "if no mode has been found in list)"),
-        NULL, 0, 0, "q:lightred;a:lightcyan;o:lightgreen;h:lightmagenta;"
+        NULL, 0, 0, "y:lightred;q:lightred;a:lightcyan;o:lightgreen;h:lightmagenta;"
         "v:yellow;*:lightblue", NULL, 0, NULL, NULL,
         &irc_config_change_color_nick_prefixes, NULL, NULL, NULL);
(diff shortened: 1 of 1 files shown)


Shawn-Smith:
I can see a possible issue with this, InspIRCd uses +y as an oper-prefix mode.

skyhighwings:
@Shawn-Smith that's not that bad, if you're oper-prefixing then you are trying to show authority, so imo that should be just fine

skyhighwings:
@Xe thoughts on above?

flashcode:
The default value in this option should work on all IRC servers. If the "y" mode has different meaning according to the server, then I think it should not be part of default value.
The option is there to customize if needed, and the default could be kept as-is for me.
Feel free to comment with different opinion or extra info, I keep this pull request opened.

Mikaela:
From what I understood +y on Elemental-IRCd is channel owner and +y on InspIRCd is oper. Both are above all others so I see no issue and @skyhighwings said so too.

I don't see any issue with this different meaning as it doesn't break anything and I don't understand why @flashcode doesn't merge it as it is.

flashcode:
Merged.

Shawn-Smith:
@Mikaela The issue is that it's a non-standard mode. Other networks could have +y as something lower than voice, or anywhere in between. +ov are the only standard prefixes, and +qah are the only widely-implemented non-standard ones. I don't think this should be added to the default.

flashcode:
It's already merged (could be reverted though).
Anyway it's just a color prefix, apart from visual look of prefix, it will not introduce any bug or cause side effects.

Mikaela:
@Shawn-Smith I don't understand your issue. Does WeeChat hardcode the importance of prefixes ignoring PREFIX from the IRCd?

PR #466, opened 2015-07-21
gui: change alt-j binds to allow jump_current_to_previous_buffer

sim642:
This patch allows using binds such as `meta-j01` exactly like `meta-1` when jumping to the current buffer by default as it is quite tedious for an user to change 100 binds.

*Feature requested by Platypus-Man at #weechat.*


flashcode:
Merged.

PR #468, opened 2015-07-27
Fix compilation problems when ncurses has split libtinfo

jlec:
Downstream bug: http://bugs.gentoo.org/507170

Try using pkg-config to detect ncurses libraries before falling back
to old mechanism.

Signed-off-by: Justin Lecher jlec@gentoo.org

### DIFF
--- a/src/gui/curses/CMakeLists.txt
+++ b/src/gui/curses/CMakeLists.txt
@@ -35,7 +35,16 @@ main.c)
 
 set(EXECUTABLE weechat)
 
-find_package(Ncurses)
+
+find_package(PkgConfig)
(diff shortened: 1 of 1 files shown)


flashcode:
Side note: same problem: https://savannah.nongnu.org/bugs/?41245

flashcode:
Hi,
Why not doing this fix in file `cmake/FindNcurses.cmake` instead?
And rename NCURSES_LIBRARY to NCURSES_LIBRARIES so that you don't need extra lib name on link.

flashcode:
Problem is resolved by the merge of #1090, so I close this issue.

PR #471, opened 2015-07-29
correct nastevil to nastavil

mimi1vx:
http://prirucka.ujc.cas.cz/?slovo=nastavil&Hledej=Hledej 

word "nastevil" in czech does not exist


flashcode:
Thanks, merged.

flashcode:
By the way if you have some time, the Czech translations need updates (they are not updated since some years).

PR #493, opened 2015-08-18
irc: add multiple targets and -server option to /ctcp (closes #204)

### LINKED ISSUE #204: Add -server option for /ctcp command
It would be nice to have a `-server` option for the `/ctcp` command, just like `/msg`, `/join` and other commands have one.

sim642:
This makes `/ctcp` syntax like `/msg`'s.

*Also includes my nickname in AUTHORS :)*


Zarthus:
Does exactly what it says on the tin.

PR #511, opened 2015-08-30
Add "recover" to the list of hidden commands.

juanfra684:
That command is used by the Blinkenshell IRC network. Similar to "ghost".


Mikaela:
It looks like this is the trigger hiding passwords in NickServ commands?

It might be worth noting that the regain command has been in Atheme for a long time and you find it almost everywhere including freenode.

You might also want to take a look at https://weechat.org/files/doc/stable/weechat_dev.en.html#contribute (the commit should be something like `trigger: add regain to msg_auth`).

juanfra684:
Thanks for the help. I updated the PR.

juanfra684:
Can anyone commit this patch? the change is pretty straightforward.

flashcode:
Sure!

PR #513, opened 2015-09-02
FindIconv: Stop hardcoding the library name.

rakuco:
Instead of splitting ${ICONV_LIBRARY}'s path and then assuming the
library is called "libiconv.so", use ${ICONV_LIBRARY}'s entire value in
the calls to check_library_exists().

If that is not done, the iconv detection fails if one sets ICONV_LIBRARY
to, for example, /usr/lib/libc.so, as the previous calls would still
pass "-liconv" to the linker instead of the chosen value.


rakuco:
ping?

flashcode:
Sorry for the long delay, I should look very soon; I have to test compilation on different OS before merging.

PR #539, opened 2015-09-25
added notes for crash report on OSX (with lldb instead of gdb)

drew1kun:
added notes for crash report on OSX (with lldb instead of gdb)


Mikaela:
I think it might be a better to just talk about "OS X" instead of specifying the release name as the updates are free and El Capitain is scheduled to be released on 30th.

PR #564, opened 2015-10-27
.travis.yml: use Ubuntu Trusty

Mikaela:
14.04 instead of 12.04 and shipping the Debian package becomes useless
as it's available in the repositories now.


flashcode:
I have run the checks again, after some fails, and there are still the same errors.
Any idea why the compilation fails with that?

    /opt/python/2.7.10/lib/python2.7/config/libpython2.7.a: error adding symbols: Bad value
    collect2: error: ld returned 1 exit status

And I don't know why some tests are OK.

flashcode:
It looks like it works only with python 3.x but not python 2.x.

flashcode:
I wonder why python is installed in `/opt/python/2.7.10` which doesn't look like a standard location.
The Trusty distro on Travis has some specific python installed?

stfnm:
Seems to me like the python2 shared library can't be found and therefore it tries linking with the static library instead.

Mikaela:
Hopefully understandable issue opened at https://github.com/travis-ci/travis-ci/issues/5021.

Mikaela:
Ready for merge with workaround on https://github.com/travis-ci/travis-ci/issues/5021#issuecomment-165618087. I wish to keep #624 open too though as it has faster tests and start time, but for now this has newer Ubuntu.

Mikaela:
Upstream issue which is worked around is https://github.com/travis-ci/travis-ci/issues/5326 and I will change it to the comment.

flashcode:
Merged, thanks.

altaurog:
How can I specify python version against which to compile when building?

altaurog:
Nevermind.  I see it in the docs: `-DENABLE_PYTHON3`

> Compile Python plugin using Python 3 (NOT recommended because many scripts are not compatible with Python 3)

(too bad!)

altaurog:
I am trying to build weechat 1.5 on Debian jessie 64 bit.  I get the same error as above:

```
Linking C shared module python.so
/usr/bin/ld: /usr/local/lib/python2.7/config/libpython2.7.a(abstract.o): relocation R_X86_64_32S against `_Py_NotImplementedStruct' can not be used when making a shared object; recompile with -fPIC
/usr/local/lib/python2.7/config/libpython2.7.a: error adding symbols: Bad value
collect2: error: ld returned 1 exit status
src/plugins/python/CMakeFiles/python.dir/build.make:112: recipe for target 'src/plugins/python/python.so' failed
make[2]: *** [src/plugins/python/python.so] Error 1
```

It is picking up one of my own python builds, which apparently was built without `--enable-shared`.  I get this same error even when I build with `PATH=/usr/bin:/bin`.  How can I instruct it to use the debian-installed python (or one of my own python builds with shared library)?

altaurog:
Just installing from the debian repo for now...

altaurog:
These are the build variables to set ([source][so]):

```
PYTHON_EXECUTABLE:FILEPATH=/usr/bin/python2.7
PYTHON_INCLUDE_PATH:PATH=/usr/include/python2.7
PYTHON_LIBRARY:FILEPATH=/usr/lib/x86_64-linux-gnu/libpython2.7.so.1.0
```

  [so]: http://stackoverflow.com/a/16045924/519015

I also had to make sure that the correct python is first on `PATH` when executing weechat in order for the python scripts to function correctly.

PR #572, opened 2015-11-03
irc: add inclusive behavior to /allchan, /allpv, /allserv

sim642:
Previously these commands could only execute commands on all channels/queries/servers except the ones specified. This patch allows the inversion of that behavior, i.e. execute commands only on specified channels/queries/servers via the `-include=` argument.


Zarthus:
When I tested around with it everything worked as expected. :)

PR #580, opened 2015-11-10
irc: add support for displaying multi-prefixes

sim642:
This isn't perfect though because all the prefixes are shown using the first one's color. Considerable changes would have to be done in the way prefixes and their colors get set to get each prefix with its respective color in the case of multiple prefixes.

Also since nicklists have no concept of alignment, nicks with multiple prefixes will make the list of nicks a bit unaligned which doesn't look that great, which is why that options is left disabled by default for now.


src/plugins/irc/irc-protocol.c
```diff
@@ -4506,15 +4520,18 @@ IRC_PROTOCOL_CALLBACK(366)
                     if (strcmp (weechat_infolist_string (infolist, "type"),
                                 "nick") == 0)
                     {
-                        ptr_option = weechat_config_get (weechat_infolist_string (infolist,
-                                                                                  "prefix_color"));
-                        length +=
-                            ((ptr_option) ? strlen (weechat_color (weechat_config_string (ptr_option))) : 0) +
```
sim642: Couldn't really figure out what this line is trying to accomplish because it differs from all the other color lengths. What difference does the option make to the length of the color returned by `weechat_color`, isn't it always the same anyway? Also, not being an option meant that no characters were allocated for prefix color which seemed weird.

If this line actually was required for something, I can leave it in as it was.

PR #592, opened 2015-11-19
trigger: only return trigger's return code when condition evaluates to true

sim642:
Previously all pieces of trigger documentation (user guide and `/help trigger`) suggested that the processing of a trigger is immediately stopped if the condition (if given) evaluates to false. This makes it seem like it should never reach the point of returning the desired return code. Actually the behavior was different and always returned the given return code.

This patch fixes that behavior to be like the documentation describes for all hook types which can return a given return code. It allows for triggers to actually conditionally return `ok_eat` because previously it wasn't possible to conditionally stop a hook from propagating.

This has the potential to break current triggers but that happening is extremely unlikely as return codes are so rarely used and if they are, they most definitely weren't conditional before. In the case that such trigger actually broke it's possible to create two triggers in the new case, one for conditionally running the command and the other unconditionally returning the return code.

### DIFF
--- a/src/plugins/trigger/trigger-callback.c
+++ b/src/plugins/trigger/trigger-callback.c
@@ -330,9 +330,13 @@ trigger_callback_run_command (struct t_trigger *trigger,
  *   2. check conditions (if false, exit)
  *   3. replace text with regex
  *   4. execute command(s)
+ *
+ * Returns:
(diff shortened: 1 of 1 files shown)


Zarthus:
Looks OK to me.

flashcode:
Rebased with necessary changes and merged, thanks!

PR #597, opened 2015-11-20
irc: indexed ban list and completion for /unban (task #11374, task #10876)

sim642:
This patch implements a new structure, **modelist**, into IRC channels. A modelist is the list of masks which are listed under a channel mode. For all server channel's, mode lists are created for every type A CHANMODE (e.g. `b`, `e`, `I`, `q`).

When receiving these lists, the structures are updated in weechat and displayed indexed, but also when getting changed afterwards the lists are kept in sync with what should be on the server. Initially a modelist is empty until it is received fully via the respective `/mode` command. After which it's possible to `/unban` or `/unquiet` by modelist indices or use completions for either command.

All additional information, such as mode mask's setter and the time, which the server sends, are also stored in the modelist, although currently unused. Also as of the moment, there's no mechanism to automatically get the modelist from server when required (such as trying to complete `/unban` before having run `/ban` to get the list).

### DIFF
--- a/doc/en/weechat_dev.en.adoc
+++ b/doc/en/weechat_dev.en.adoc
@@ -232,6 +232,7 @@ WeeChat "core" is located in following directories:
 |       irc-input.c                 | Input of commands/text.
 |       irc-message.c               | Functions to manipulate IRC messages.
 |       irc-mode.c                  | Functions about channel/nick modes.
+|       irc-modelist.c              | IRC channel mode lists (+b, +e, +I)
 |       irc-msgbuffer.c             | Target buffer for IRC messages.
 |       irc-nick.c                  | IRC nicks.
 |       irc-notify.c                | IRC notify lists.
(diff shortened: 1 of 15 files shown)


Zarthus:
As written in the description, when I used this I noticed I had to manually `/mode #channel b` to retrieve autocompletion for unbanning, otherwise it would cycle through nicknames. Do you think it would be possible to follow up on this by polling `/mode #channel b` when using tab in `/unban`? (or to have weechat poll banlists automatically, but I think this is outside of the scope of the pull request).

All in all, in its current state, it works as the description describes it. If merged and used in combination with issue 579, I would become a very happy man.

sim642:
@Zarthus I've thought about the automatic ban list requesting when trying to complete `/unban` but I'm not exactly sure how WeeChat does the completion handling in the sense that, if new completions were to appear in the process of trying to complete it already, would anything change or will it apply next time when doing the autocompletion. An extra thing that I could have a look at sometime.

sim642:
```
< opal>    sim642 you made a really simple mistake                           
< opal>    you assumed setter would be non null in irc_modelist_item_new     
< opal>    test your patch on ircnet                                         
< opal>    null pointer deref on line 86                                     
```

```
< opal>    http://git.volatile.ch/krustykrab/neptune/commits/modelist  
< opal>    theres the branch im using                                  
```

PR #604, opened 2015-11-26
Fixed instructions about getting correct CA file on Mac OS X

kron4eg:
curl-ca-bundle was removed from Homebrew as insecure: https://github.com/Homebrew/homebrew/pull/28658 so here is the alternative version of getting correct and secure CA file.


flashcode:
Merged, thanks!

PR #611, opened 2015-11-30
core: add wildcard matching operators to eval (closes #608)

### LINKED ISSUE #608: Eval operator for matching wildcard/list pattern
Some options like `weechat.plugin.autoload` and function `weechat_buffer_match_list` in plugin/scripting API accept comma separated list of entries that can contain wildcard (`*`) and negation (`!`). It would be great if we can use this kind of pattern in eval's condition because it is nicer and much easier to understand than posix regex. For example, here's a condition for matching user host:

```
${host} =% bot@example.com,*@do.ma.in.*,!troll@do.ma.in.*,*@weechat.org
```

_(the `=%` is just dummy placeholder for this not-yet-exist operator)_

with the current eval it would look like this:

```
${host} == bot@example.com || (${host} =~ .*@do\.ma\.in\..* && ${host} !~ ^troll@do\.ma\.in\..*) || ${host} =~ .*@weechat\.org$
```

I'm sure some people could rewrite the above condition into one posix regex pattern but most people couldn't.

The other advantage of this pattern is you don't have to escape a lot of characters so it's good for matching against list of IRC-related stuffs like nicks and channels.

sim642:
This patch adds two operators, `=*` and `!*`, for matching/not matching strings by wildcard mask instead of more complicated regex. The `=*` operator was chosen instead of the suggested `=%` because the latter has lesser connection with wildcards, in my opinion.

Also this patch does not implement the "match against a list" behavior asked for because it does not make sense for one operator to be able to match against a list, while all the others match against a single string.


src/core/wee-eval.h
```diff
@@ -42,6 +42,8 @@ enum t_eval_comparison
 {
     EVAL_COMPARE_REGEX_MATCHING = 0,
     EVAL_COMPARE_REGEX_NOT_MATCHING,
+    EVAL_COMPARE_MATCHING,
```
flashcode: For less ambiguity, I would use name `EVAL_COMPARE_STRING_MATCHING` and `EVAL_COMPARE_STRING_NOT_MATCHING`.

flashcode:
If possible, could you please add the tests of the new operators in tests/unit/core/test-eval.cpp ?
If you're not sure, I can do it after the merge, not a problem.

src/core/wee-eval.h
sim642: I thought about naming it better too, I'll modify that then.

sim642:
I can also add some tests, sure. Forgot because I don't usually add features to the core.

flashcode:
Thanks, merged.

PR #613, opened 2015-11-30
Added irc.server specifc ssl_ca_file config option

### LINKED ISSUE #438: Allow to specify ca certificates per network
I would like to connect to the hackint IRC network via TLS. They use their own root CA which I would like to specify only for the hackint network: http://hackint.eu/
However this does not seem to be possible, since there seems to be only a globel CA certificate setting. But setting it to the hackint CA would break accessing the Freenote IRC network. Therefore please allow to set the CA per IRC network additionally to the global default setting.

ManiacTwister:
Added ability to specify gnutls ca_file per server. This fixes  #438

### DIFF
--- a/src/plugins/irc/irc-config.c
+++ b/src/plugins/irc/irc-config.c
@@ -1716,7 +1716,8 @@ irc_config_server_new_option (struct t_config_file *config_file,
             new_option = weechat_config_new_option (
                 config_file, section,
                 option_name, "string",
-                N_("File containing the certificate authorities. "),
+                N_("file containing the certificate authorities (\"%h\" will be "
+                   "replaced by WeeChat home, \"~/.weechat\" by default)"),
                 NULL, 0, 0,
                 default_value, value,
                 null_value_allowed,
--- a/src/plugins/irc/irc-config.c
+++ b/src/plugins/irc/irc-config.c
@@ -1712,6 +1712,18 @@ irc_config_server_new_option (struct t_config_file *config_file,
                 callback_change, callback_change_data,
                 NULL, NULL);
             break;
+        case IRC_SERVER_OPTION_SSL_CA_FILE:
+            new_option = weechat_config_new_option (
+                config_file, section,
+                option_name, "string",
+                N_("File containing the certificate authorities. "),
+                NULL, 0, 0,
+                default_value, value,
+                null_value_allowed,
+                callback_check_value, callback_check_value_data,
+                callback_change, callback_change_data,
+                NULL, NULL);
+            break;
         case IRC_SERVER_OPTION_PASSWORD:
             new_option = weechat_config_new_option (
                 config_file, section,
--- a/src/plugins/irc/irc-server.c
+++ b/src/plugins/irc/irc-server.c
@@ -4053,6 +4053,11 @@ irc_server_gnutls_callback (void *data, gnutls_session_t tls_session,
 
             if (ca_file2)
             {
+                weechat_printf (
+                    server->buffer,
+                    _("%sgnutls: set ca_file to %s"),
+                    weechat_prefix ("network"),
+                    ca_file2);
                 gnutls_certificate_allocate_credentials (&server->gnutls_xcred);
                 gnutls_certificate_set_x509_trust_file (server->gnutls_xcred, ca_file2,
                                                                 GNUTLS_X509_FMT_PEM);
--- a/src/plugins/irc/irc-server.c
+++ b/src/plugins/irc/irc-server.c
@@ -3203,6 +3203,7 @@ irc_server_close_connection (struct t_irc_server *server)
             if(server->gnutls_xcred != NULL) {
                 gnutls_credentials_clear(server->gnutls_sess);
                 gnutls_certificate_free_credentials(server->gnutls_xcred);
+                server->gnutls_xcred = NULL;
             }
 
             gnutls_deinit (server->gnutls_sess);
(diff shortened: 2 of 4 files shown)


jvvv:
I've tested this with both stable and current. For both cases, I often get sigabort when disconnecting from a server. I've pasted a backtrace here: http://sprunge.us/aQAI

ManiacTwister:
@jvvv Thanks, fix pushed

jvvv:
I apologize for taking so long to come back to this. Fix works for me; no more sigaborts. On another note, I notice a difference in connecting. With your changes, the client doesn't login automatically any more. I tried building in the following was to compare results:
1) vanilla 1.3 release
2) 1.3 release, patched for ssl_ca_file feature
3) vanilla git HEAD
4) git HEAD patched for ssl_ca_file feature
5) direct clone from your fork

All cases that use the ssl_ca_file feature fail to login at connect. Vanilla 1.3 and vanilla git HEAD work as expected. Still able login via <code>/msg nickserv identify</code>. I haven't had time to investigate further.

ManiacTwister:
@jvvv I can reproduce this. Obviously the ```irc_server_gnutls_callback``` hook is never called with the ```WEECHAT_HOOK_CONNECT_GNUTLS_CB_SET_CERT``` action if the ca file is set. I'll dig into this issue deeper tomorrow.

ManiacTwister:
I think we have to add a  ``` gnutls_certificate_credentials_t gnutls_xcred ``` argument the following three functions (or duplicate the functions into irc-server.c):
```
network_set_gnutls_ca_file
network_init_gnutls
network_end
```
Then we need to call it with the general network xcred and again with the server xcred if applicable.

src/plugins/irc/irc-config.c
```diff
@@ -1712,6 +1712,18 @@ irc_config_server_new_option (struct t_config_file *config_file,
                 callback_change, callback_change_data,
                 NULL, NULL);
             break;
+        case IRC_SERVER_OPTION_SSL_CA_FILE:
+            new_option = weechat_config_new_option (
+                config_file, section,
+                option_name, "string",
+                N_("File containing the certificate authorities. "),
```
sim642: For consistency this option description should be similar to the others, i.e. start lowercase and no full stop.

src/plugins/irc/irc-server.c
```diff
@@ -63,6 +63,7 @@
 #include "irc-raw.h"
 #include "irc-redirect.h"
 #include "irc-sasl.h"
+#include "../../core/wee-hook.h"
```
sim642: Weechat plugins (in this case "irc") shouldn't directly interact with the core, that's what the plugin API is for.

src/plugins/irc/irc-server.c
```diff
@@ -4032,6 +4042,51 @@ irc_server_gnutls_callback (void *data, gnutls_session_t tls_session,
         fingerprint = IRC_SERVER_OPTION_STRING (server,
                                                 IRC_SERVER_OPTION_SSL_FINGERPRINT);
 
+        /* get ca_file option in server */
+        ca_file0 = (char *) IRC_SERVER_OPTION_STRING (server,
+                                                IRC_SERVER_OPTION_SSL_CA_FILE);
```
sim642: Weird indentation.

src/plugins/irc/irc-server.c
```diff
@@ -4032,6 +4042,51 @@ irc_server_gnutls_callback (void *data, gnutls_session_t tls_session,
...
+            if (ca_file2)
+            {
+                weechat_printf (
+                    server->buffer,
+                    _("%sgnutls: set ca_file to %s"),
+                    weechat_prefix ("network"),
+                    ca_file2);
+                gnutls_certificate_allocate_credentials (&server->gnutls_xcred);
+                gnutls_certificate_set_x509_trust_file (server->gnutls_xcred, ca_file2,
+                                                                GNUTLS_X509_FMT_PEM);
```
sim642: Weird indentation.

src/plugins/irc/irc-server.c
```diff
@@ -4032,6 +4042,51 @@ irc_server_gnutls_callback (void *data, gnutls_session_t tls_session,
...
+                gnutls_credentials_set(tls_session, GNUTLS_CRD_CERTIFICATE, server->gnutls_xcred);
```
sim642: Missing space after function name.

src/plugins/irc/irc-server.c
```diff
@@ -4032,6 +4042,51 @@ irc_server_gnutls_callback (void *data, gnutls_session_t tls_session,
...
+
+                /*
+                 * We need to set verify/retrieve functions for our server
+                 * specific gnutls_xcred here.
+                 */
+                #if LIBGNUTLS_VERSION_NUMBER >= 0x02090a /* 2.9.10 */
```
sim642: Preprocessor statements shouldn't be indented, regardless of their nested position.

src/plugins/irc/irc-server.c
```diff
@@ -4332,6 +4387,10 @@ irc_server_gnutls_callback (void *data, gnutls_session_t tls_session,
     if (cert_temp_init)
         gnutls_x509_crt_deinit (cert_temp);
 
+/*    if(gnutls_xcred_init) {
+        gnutls_credentials_clear(tls_session);
+        gnutls_certificate_free_credentials (gnutls_xcred);
+    }*/
```
sim642: Committing comments is useless.

sim642:
It looks to me like this patch wants to use `hook_connect_gnutls_set_certificates` from the core and does that in an unacceptable (including core directly, not using the API) and broken (autotools builds fail due to linker error) way. If it really needs to work in that way it'd be better if the respective functionality was moved into core to begin with, allowing any SSL connection to specify its own CA file, not just IRC servers.

ManiacTwister:
I'll look on how to implement it without including the core soon.

flashcode:
Closing in favor of #1262.

PR #615, opened 2015-12-01
irc: make command characters optional in server's command option

sim642:
The motivation for this patch came from aliases which allow specifying commands with and without the command character. Server `.command` variables could easily do the same without negative consequences because previously without the command character it would simply attempt to send text into the server buffer, only to end up with an error message.


KamranMackey:
Please update the Changelog in this PR, that's why I can't apply the commit to my local repo. The ChangeLog isn't up-to-date in that branch.

flashcode:
I think the ChangeLog should not be updated in pull requests, I can do that after (updating it will often lead to conflits with current master branch).

PR #616, opened 2015-12-02
trigger: allow regex delimiter in regex if escaped

sim642:
Previously using a regex like `/^@/\/me /` did not work for triggers, due to invalid regex format. When splitting the regex by delimiter, the delimiter was found directly with `strstr`, not checking if it was escaped or not. This patch allows using the delimiter character within the regex if it is properly escaped.

The function `trigger_convert_escaped_chars` is implemented similarly to `string_convert_escaped_chars` but leaves all other escapes other than the delimiter's untouched. It then passes the result through `string_convert_escaped_chars` to handle the rest of the escapes. The complexity of this function (as opposed to just using `string_replace` for the escaped delimiter) is required to properly handle the corner cases where the order of applying the two functions changes the result compared to doing the escapes at the same time.

### DIFF
--- a/src/plugins/trigger/trigger.c
+++ b/src/plugins/trigger/trigger.c
@@ -483,6 +483,62 @@ trigger_regex_free (int *regex_count, struct t_trigger_regex **regex)
 }
 
 /*
+ * Converts escaped chars to their value, including the delimiter.
+ *
+ * Note: result must be freed after use.
+ */
+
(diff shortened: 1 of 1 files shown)


arza-zara (commented):
Seems to break matching \ in hook command, which is probably a core issue (#729). Otherwise seems fine.

Issue #559, opened 2015-10-23
Alt-Backspace should eat less than Ctrl-W


purpleidea:
In pretty much every terminal environment I know, including bash and irssi, pressing alt-backspace eats one word at a time. For example:

```
word1.word2-word3_word4
```

Would get "eaten" by four repeated presses. Unfortunately, in weechat, all four disappear with a single stroke. The 1.2 release notes allude to this being configurable:

`"- add options to customize word chars (for detecting word boundaries)"` from https://weechat.org/news/87/20150510-Version-1.2/

but if that's the case, I haven't figured out how. I'd like to request this functionality, and if it is available and someone can explain how to enable it, then perhaps a docs patch is in order.

Thanks in advance!

flashcode:
Hi,

Yes, since WeeChat 1.2, you can customize the word chars, and by default with your example the key Ctrl-W should remove the three last words, then the first one. Which is not exactly what you want.
If you want to consider that `-` and `_` are NOT part of words, then you can do:

    /set weechat.look.word_chars_input  "!\u00A0,|,alnum"

Then Ctrl-W should delete words one by one.

purpleidea:
@flashcode Thanks for the response however, Ctrl-W should delete the whole chunk, where as alt-backspace should delete one word at a time.

Is there a way to set this behaviour?

flashcode:
No, in WeeChat Ctrl-W and Alt-backspace are bound on same command. And there's only one command to remove the previous word, so no way to have a different behavior for both keys.

flashcode:
I just tried in bash, and I see the difference. Then I turn this issue as an feature request, to introduce different behavior between Ctrl-W and Alt-backspace.

purpleidea:
On Fri, Oct 23, 2015 at 1:51 AM, Sébastien Helleu <notifications@github.com>
wrote:

> I just tried in bash, and I see the difference. Then I turn this issue as
> an feature request, to introduce different behavior between Ctrl-W and
> Alt-backspace.



Much appreciated thanks. Maybe you can steal the necessary code from irssi
:)

Cheers!

flashcode:
Stealing code is not an option for me ;)

purpleidea:
@flashcode Come disait Picasso: "Good artists copy, great artists steal."

Les licenses sont compatible :)

andreaskern:
Ctrl-W should delete to the previous whitespace
Meta-Del should delete to the previous word boundary

[...]

PR #622, opened 2015-12-13
Add new /input delete_backwards_until_space that acts like alt+backspace, not ctrl+w

FauxFaux:
Re-use most of the logic from `/input delete_previous_word` to make a new command that stops at spaces, like meta+backspace does in readline applications.  See #559 for more discussion.

I actually contest that ctrl+w behaves like the issue claims anywhere, and that this weechat behaviour is weird and wrong, but it doesn't seem that harmful to have both behaviours available as different commands.

Could easily be implemented without the extract methods, but I didn't think it could and I don't think the extract methods make it worse.

### DIFF
--- a/ChangeLog.adoc
+++ b/ChangeLog.adoc
@@ -36,7 +36,7 @@ Bug fixes::
   * relay: fix send of "PART" command in backlog (irc protocol)
   * relay: fix parsing of CAP command without arguments in irc protocol, send ACK only if all capabilities received are OK and NAK otherwise (issue #1040)
 
-Tests:
+Tests::
 
   * display an error if the required locale en_US.UTF-8 is not installed
 


rr-:
>I actually contest that ctrl+w behaves like the issue claims anywhere, and that this weechat behaviour is weird and wrong

I switched from irssi today and I agree that the current implementation of c-w is very annoying. The fix from #559 mitigates it to some extent but it still has some issues e.g. when referencing other people:

1. `nick: word|`
2. *ctrl+w*
3. `nick:|` instead of `nick: |`

yet-another-random-username:
I think the hassle here is that it deletes more than just the word. It deletes the word and all non-word characters bar the first one. Since colon is considered a non-word char as much as a space you get the cursor next to the colon. It'd be nice if it stopped at the first non-word char, instead.

This would have other benefits. At the moment this happens: `http://www` -> ^w -> `http:`

If it stopped at the first non-word char, this would happen: `http://www` -> ^w -> `http://`

The second behaviour would allow you to just retype with the correct hostname. The first (current) behaviour does not.

rr-:
Why?

FauxFaux:
Github has interpreted me rearranging my branches (via. force pushes) as closing the pull request.

The original 1.7 branch is: https://github.com/FauxFaux/weechat/tree/faux-1.7
The 1.9 based rework is in master.

It's been two and a half years; maybe it's time to admit that it isn't going to get merged. (I'm still running my own fork.)

nyuszika7h:
Yeah, you can't rename a pull request branch, you have to open a new PR in
that case.

flashcode:
For your information, #559 was implemented: now the new key Alt+backspace eats less than Ctrl+w (delete previous word until whitespace).

Issue #625, opened 2015-12-16
regular expressions in weechat do not (entirely) support unicode


oakkitten:
>> オ =~ [アオ]
    == [1]
    >> オ =~ [ア-オ]
    == [0]

flashcode:
I'm not sure if the problem is in WeeChat itself, I think it could be in POSIX regex support. If that's the case, this would end as "won't fix", except if the purpose of the issue is to change the regex support in WeeChat (use another regex like pcre).

oakkitten:
people on #regex don't know whether or not posix supports unicode. but i tried https://github.com/sim642/weechat/tree/pcre and that didn't work either.

oakkitten:
@sim642's pcre branch doesn't solve the problem, either
in the pcre branch,

    >> ф =~ р
    == [0]
    >> ф =~ [р]
    == [1]
    >> ф =~ [п]
    == [0]

`ф`, `п` and `р` are, respectively, `\u0444`, `\u043f` and `\u0440`, which is `\xd1\x84`, `\xd0\xbf and `\xd1\x80` in utf-8.

as for posix, i don't really know if it supports unicode and #regex doesn't know either :s

sim642:
From @oakkitten's last example (which I suspect works the same in current POSIX regex) it seems that the regex library does not treat the utf8 encoded characters as characters but simply looks at the ascii characters (bytes). That makes the second condition true because `\xd1\x84` contains a character from the **two**-character class `[\xd1\x80]`.

It is, however, odd that some unicode has seemed to work (like chinese characters).

If I find the time I can have a deeper look into my pcre branch and see if there needs to be something extra enabled to get full unicode support with pcre.

sim642:
With the use of pcre's REG_UTF8 flag I've managed to make it work like this:
```
>> オ =~ [アオ]
== [1]
>> オ =~ [ア-オ]
== [1]
>> ф =~ р
== [0]
>> ф =~ [р]
== [0]
>> ф =~ [п]
== [0]
```
which all seem to be correct.

This gives actual reason to get my pcre branch ready for a PR.

stfnm:
I'd enjoy PCRE support too. :+1:

oakkitten:
a note on the current behavior of highlight regex matching: `sq|(sq|ab)\w` matches `sq`, `abc` but not `sqr`. `(sq|ab)\w|sq` matches all 3. i think i understand the mechanism, but it's not very obvious. maybe it could be made a bit more obvious.

PR #627, opened 2015-12-20
core: add PCRE support [WIP]

sim642:
This patch adds support for PCRE instead of POSIX regex via pcreposix wrapper. 

Build option `ENABLE_PCRE` needs to be enabled to compile with PCRE support (default `OFF`). Since I'm not too familiar with autotools, there's no option for it with that but should be added in the future.

PCRE allows proper UTF-8 in regular expressions (#625). Besides all the additional regex features it's also possible to use named captures in `eval_expression`'s `regex_replace`.


sim642:
PCRE completely fails the tests because the regex flag handling works a bit differently. It needs to be figured out what to do about this problem.

flashcode:
I think you could have conditional code in tests to have different tests with POSIX regex and PCRE.

sim642:
I know I could but that basically means ditching most of the existing regex tests with PCRE as `string_regex_flags` is only needed to force some flags for pcreposix, while letting PCRE do the parsing of flags in the pattern itself. Somewhat defeats the purpose of the tests.

DemiMarie:
I think we should use TCL's regex engine instead of PCRE, or provide both.  TCL's regex engine is faster and immune to catastrophic backtracking.

sim642:
Is TCL's regex engine available as a separate C library? Does it act as a drop-in replacement for POSIX? These are two fundamental features that PCRE has which makes the migration straightforward. Also PCRE has specifically been requested repeatedly and is more widely known (this is the first time I hear about TCL's).

According to benchmarks PCRE should already be faster than POSIX so that's a step up for us. 

Providing both would mean having three different regex engines all of which would have to be supported. It's also likely to introduce incompatibilities between users using different engines.

DemiMarie:
I think we should use TCL's regex engine instead of PCRE, or provide both.  TCL's regex engine is faster and immune to catastrophic backtracking.

PR #635, opened 2015-12-25
core: add hash seed to nick colors to allow for reshuffling of colors

sim642:
This patch adds the possibility of shuffling nick colors without having to choose a different hash algorithm or adding/removing colors in `chat_nick_colors`. The seed option's string is prepended to the nick before it's hashed. The default of an empty string means that by default, nothing gets changed from previous versions.

_Feature requested by m_ben in #weechat._

### DIFF
--- a/src/core/wee-config.c
+++ b/src/core/wee-config.c
@@ -152,6 +152,7 @@ struct t_config_option *config_look_mouse;
 struct t_config_option *config_look_mouse_timer_delay;
 struct t_config_option *config_look_nick_color_force;
 struct t_config_option *config_look_nick_color_hash;
+struct t_config_option *config_look_nick_color_hash_seed;
 struct t_config_option *config_look_nick_color_stop_chars;
 struct t_config_option *config_look_nick_prefix;
 struct t_config_option *config_look_nick_suffix;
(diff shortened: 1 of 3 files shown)


flashcode:
If you rebase your commits on top of current master, tests should be OK.

Zarthus:
Does exactly what it says on the tin, I've found no problems when I tested this.

src/core/wee-config.c
```diff
@@ -152,6 +152,7 @@ struct t_config_option *config_look_mouse;
 struct t_config_option *config_look_mouse_timer_delay;
 struct t_config_option *config_look_nick_color_force;
 struct t_config_option *config_look_nick_color_hash;
+struct t_config_option *config_look_nick_color_hash_seed;
```
sim642: `salt` instead of `seed` would be a more appropriate name: the value is used as a salt to a hash function, not as a seed to some PRNG.

src/gui/gui-nick.c
```diff
@@ -57,7 +59,21 @@ gui_nick_hash_color (const char *nickname)
     if (config_num_nick_colors == 0)
         return 0;
 
-    ptr_nick = nickname;
+    length = strlen (CONFIG_STRING(config_look_nick_color_hash_seed)) +
```
sim642: Add conditional to only do the `malloc` and salt concatenation when the salt is non-empty. Avoids unnecessary allocations and string operations by default and in most cases.

sim642:
For future reference, there are further micro-optimizations that could be done but aren't necessary:

1. Instead of allocating a new string and concatenating with the salt as prefix, it's possible to avoid the allocation. That requires duplicating each hash function implementation to first iterate over the salt characters and then a second loop over the nick characters. This works because all current hash functions work by iterating through the characters once. The downside of this is uglier and duplicate code.

2. Instead of recalculating the hash function of the salt characters each time, like described above, that value could be cached and reused as the initial value for just iterating over the nick characters. This works because all current hash functions start from a constant value. It would be important to recalculate the cached initial value not only when the seed is changed but also when the hash function is changed.

flashcode:
For the first remark, I'll update the function to not allocate a new string, and using a double loop in hash algorithm (no duplicate code), to loop on both strings (if salt is set to a non-empty string).

For the second remark, it could be done, but I think this is not a so critical code, and not necessary to to such optimization.

flashcode:
Merged, thanks.

PR #656, opened 2016-01-15
Add extern keyword several places to link in osx

mojadita:
This was causing trouble in osx linking

    duplicate symbol _trigger_buffer in:
        .libs/trigger.o
        .libs/trigger-buffer.o

and others (same error)


flashcode:
Thanks for the fix (and weird that it caused troubles on OS X only).

PR #659, opened 2016-01-18
Another non-extern bad global declaration.

mojadita:
Se pull request #656 

another non-extern declared global variable leads to multiple defined symbols on linking.  Perhaps some dependencies for header files must be checked?


flashcode:
These changes have already been merged by #656, so I close this one.

PR #660, opened 2016-01-18
Encode decode from server is misordered

mojadita:
the charset encoder/decoder is broken.  When encoding to server, we first construct the message and just before sending we encode it, so all masks and channels are encoded in weechat in utf-8 and to the server in its charset.  On receiving, messages are first parsed and then partially decoded, making the masks and targets to remain in the native charset of the server, which is incorrect.  I have patched the receiving code to allow for server charset masks (irc-hispano allows to fix the domain and allows iso-8859-15 chars in virtual addresses, also in channel names)  making unnecessary the config parameter **irc.network.channel_encode** and allowing only the `irc.<server>` modifier (no more allowance to `irc.<server>.<channel_or_nick>` modifiers in charset decoding)

### DIFF
--- a/src/plugins/irc/irc-config.c
+++ b/src/plugins/irc/irc-config.c
@@ -134,7 +134,6 @@ struct t_config_option *irc_config_color_topic_old;
 struct t_config_option *irc_config_network_autoreconnect_delay_growing;
 struct t_config_option *irc_config_network_autoreconnect_delay_max;
 struct t_config_option *irc_config_network_ban_mask_default;
-struct t_config_option *irc_config_network_channel_encode;
 struct t_config_option *irc_config_network_colors_receive;
 struct t_config_option *irc_config_network_colors_send;
 struct t_config_option *irc_config_network_lag_check;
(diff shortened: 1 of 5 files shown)


src/plugins/trigger/trigger-buffer.h
```diff
@@ -22,7 +22,7 @@
 
 #define TRIGGER_BUFFER_NAME "monitor"
 
-struct t_gui_buffer *trigger_buffer;
```
flashcode: This has already been merged by #656.

mojadita: Yes, it's undone in a further commit in this same branch.

flashcode: Your undo is on file xfer-config.h but not this file (trigger-buffer.h).
And here I'm commenting on the "Files changed" tab in pull request, so this includes all your commits.

mojadita: Probably my branch launches from a previous commit to the one that incorporates those changes.  Let me revise it this evening, as now I'm at work.

mojadita:
I have corrected the problems identified this morning.  I think the branch is ok now (I have rebased it on top of master)

mojadita:
Please, abandon and try the new pull request on a rebased branch #661

PR #662, opened 2016-01-20
display local server names for ctcp requests

tomboy-64:
This patch changes the message appearing when weechat notices ctcp requests, so it includes info about the used IRC network.

For example:
```17:46 -- CTCP requested by Blob @ freenode: PING```
instead of
```17:46 -- CTCP requested by Blob: PING```


flashcode:
Discussed on #weechat: you can force highlight on CTCP to get notified if needed, with this command: `/set weechat.look.highlight_tags irc_ctcp` (with that, it should not be needed to redirect CTCP messages to the current buffer, which is not a good idea for me).

PR #665, opened 2016-01-23
xfer: Compute speed/eta with microsecond precision

talisein:
Am trying to get rid of the sometimes-reported 2592352 TB/s (or some other enormous value) when just starting a file xfer.

It also seems to make the periodic updates a little more stable, but maybe I am imagining that.


flashcode:
Finally merged, sorry for the delay! 😄

PR #666, opened 2016-01-23
api: add function list_user_data()

talisein:


talisein:
Updated commit to refer to 1.6 instead of 1.5. Also GPG signed for sparkly github button, ooo.

flashcode:
Thanks, merged with some changes for version 2.6.

Issue #673, opened 2016-01-27
Force a color for a nick on a specific server


jazzpi:
Would be nice to be able to have something like `/set irc.look.nick_color_force` restricted to a specific server (or maybe even channel) only.

wohali:
I know there's a 2-year old WIP for this, but the ability to `/set irc.look.{server_default|server}.nick.color` would be amazing. That syntax would allow the same nick across multiple servers to always be coloured the same.

PR #674, opened 2016-01-29
irc: more powerful nick coloring support [WIP]

### LINKED ISSUE #168: Support colouring nicks by mode (@+) and possibly hostmask.
Hi, I've used mIRC for years and one feature I really liked is colouring names by mode. This adds consistency and gives a visual effect that makes it really easy to determine who is what.

I've added images just so it is clear how it works in mIRC. While I don't expect the same setup to be copied, I think the extra context is nice.

[How my setup works in mIRC](http://i.imgur.com/U3WcYvY.png), and [how it looks in production](http://i.imgur.com/ADYlgZL.png)

This colours both the names in the nicklist, and in general chat.

It could be as simple as 

```
/set weechat.color.use_modecolor "yes"
/set weechat.color.ops "lightmagenta"
/set weechat.color.voices "green"
/set weechat.color.none "gray"
```

I personally am not that fond of the random colours.

sim642:
Changes made by this PR:
1. **New hook type "provider" based on "modifier"** (functions `hook_provider` and `hook_provider_exec`).
   
   When executing modifiers they are all applied one after another until they've all been applied (or any one of them along the line returns an empty string).
   
   Providers are very similar to modifiers based on their arguments but are called in a different way: executing providers they are tried until one returns a non-empty string. If the provider does not want to return any value based on the arguments given, it returns an empty string and next providers are executed. The order of providers is specified by their hook priorities. While modifiers work on `char *`, providers return `const char *`.
2. **Find IRC nick colors using provider hook** (provider `irc_nick_color`).
   
   Previously `irc_nick_find_color` and `irc_nick_find_color_name` functions were called internally and it was impossible for other plugins or scripts to modify the behavior of color picking. Making the two functions execute a new `irc_nick_color` provider allows scripts to create provider hooks that optionally return a color name which they'd like the nick to have, different from the one WeeChat has given it.
   
   The two existing nick coloring methods have been turned into provider hooks:
   1. Forced nick color (option `irc.look.nick_color_force`) is provided with priority 5000.
   2. Hashed nick color is provided with priority 1000.
3. **Pass server and channel to IRC nick color providers**. _Lays foundation for #673._
   
   Previously IRC nick colors were global and couldn't be different for the same nick in different servers or channels. Now all calls to get nick color in "irc" plugin pass server information (and channel information if available) to the nick coloring routines. Provider `irc_nick_color` is now being passed `provider_data` in the form `[server[;channel]]`, allowing hooks to base their choices based on those.
   
   Infos `irc_nick_color` and `irc_nick_color_name` may now include optional server and channel name in arguments to pass that info to `irc_nick_color` providers. The argument format for the infos is now `[server,[channel,]]nickname`.
4. **Allow forcing IRC nick colors based on channel mode char** (option `irc.look.nick_color_force_modes`). _Closes #168._
   
   Cherry-pick of sim642/weechat@384c8b3ae24fa09d1884a9416ce4d593379c279f modified to be a `hook_provider` and use mode chars instead of prefixes. This fixes the original commit's problem of being inconsistent with `irc.color.nick_prefixes`.
5. **Allow forcing refresh of cached IRC nick colors** (signal `irc_nick_refresh_color`).
   
   When `irc_nick_color` providers decide to color a nick differently at some point the change won't be immediately made in cached contexts, such as `struct t_irc_nick`'s `color` property, meaning some colors will be wrong. To fix this it's possible to call the signal `irc_nick_refresh_color` with arguments in the form `[server[;channel[;nick]]]` to force the refresh the color in the structures for all/server/channel/nick.

---

Missing changes:
1. ~~**Availability of provider hooks in scripts.**~~
   
   While `hook_provider` and `hook_provider_exec` are available for plugins to use, they should also be exposed to scripts because scripts are the main point where custom nick coloring may be wanted. For example, a twitch.tv script would like to have nicks colored the same way the server tells it to, not how WeeChat decides.


doc/en/weechat_plugin_api.en.asciidoc
```diff
@@ -9722,6 +9756,125 @@ weechat.hook_modifier_exec(modifier, modifier_data, string)
...
+* pointer to new hook, NULL if error occurred
+
+List of providers used by WeeChat and plugins:
+
+[width="100%",cols="^2,3,4,4",options="header"]
+|===
+| provider | provider data | String | Output
+
+| irc_nick_color |
+  |
```
sim642: Missing info about `provider_data` for `irc_nick_color`.

wohali:
@sim642 is this work abandoned? Do you need any help?

sim642:
As the [WIP] indicates, this isn't really merge-ready stuff. As the activity indicates, no work has been done on it either. Unfortunately so much time has passed that I don't really remember the exact problems that there still are but I remember there being something wrong in this approach. Here's some guesses:
1. It's not entirely clear if a whole new API hook group (providers) is needed to really implement something like this and modifiers wouldn't really do. The whole `const`-ness thing is a bit fuzzy.
2. There is a significant performance overhead and penalty with having to execute scripts and triggers as opposed to calling a simple native C function like currently.
3. There might be still difficulties with the whole nick color refreshing thing because if it depends on external information and logic from scripts, Weechat core cannot know these things automatically.
4. AFAIK since I worked on it, certain nick coloring logic has been abstracted in Weechat from IRC plugin to the core, which requires reworking this also.
5. I think the main issue is something regarding allocation and freeing of color code strings or such. Ownership and responsibility of this data might need to be reworked in Weechat to completely get around this. Currently certain uses might not be safely and correctly possible with this.

The above list is purely speculative and just a list of random thoughts that popped into my mind right now. While this would be a powerful feature to have and this PR might seem like it delivers, there's still a lot of work to be done before anything like this can be considered production ready. Feel free to investigate these changes yourself and share any thoughts.

I am not really actively working on any Weechat PRs recently anyway

PR #680, opened 2016-02-05
xfer: ignore signals when polling socket during file receive (closes #677)

### LINKED ISSUE #677: DCC transfer failed when resizing a terminal emulator window
Hello,

I have detected a weird issue, it seems that when you resize a terminal emulator window like xterm or urxvt with weechat in it, all DCC transfer that was active, will now be in 'failed' status.

I have reproduced this issues on:

* Linux (Archlinux)
* i3-wm as Windows Manager
* xterm and rxvt-unicode as terminal emulators

Weechat versions concerned:

* WeeChat 1.4 [compiled on Feb  2 2016 09:40:40]
* WeeChat 1.5-dev (git: v1.4-32-g02bcee0) [compiled on Feb  3 2016 17:13:51]

Step to reproduce:

* Launch a DCC transfer to you weechat client
* Change the dimension of your terminal emulator (with or without being in the 'xfer.list' buffer, it seems that it's not related!)

Thank you for your time,
hasB4K

sim642:
When signals (e.g. `SIGWINCH` for terminal resize) are fired they cause poll to fail with `EINTR`, erroring file receival even when there wasn't a problem with it.

This patch adds additional checks for `EINTR` and `EAGAIN` that cause retry of poll, since both are unrelated to actual file receival.

### DIFF
--- a/src/plugins/xfer/xfer-dcc.c
+++ b/src/plugins/xfer/xfer-dcc.c
@@ -368,9 +368,14 @@ xfer_dcc_recv_file_child (struct t_xfer *xfer)
         ready = poll (&poll_fd, 1, -1);
         if (ready <= 0)
         {
-            xfer_network_write_pipe (xfer, XFER_STATUS_FAILED,
-                                     XFER_ERROR_RECV_BLOCK);
-            return;
+            if ((errno == EINTR) || (errno == EAGAIN))
+                continue;
+            else
+            {
+                xfer_network_write_pipe (xfer, XFER_STATUS_FAILED,
+                                         XFER_ERROR_RECV_BLOCK);
+                return;
+            }
         }
 
         /* read maximum data on socket (until nothing is available) */


src/plugins/xfer/xfer-dcc.c
```diff
@@ -368,9 +368,14 @@ xfer_dcc_recv_file_child (struct t_xfer *xfer)
         ready = poll (&poll_fd, 1, -1);
         if (ready <= 0)
         {
-            xfer_network_write_pipe (xfer, XFER_STATUS_FAILED,
-                                     XFER_ERROR_RECV_BLOCK);
-            return;
+            if ((errno == EINTR) || (errno == EAGAIN))
```
talisein: Seems like xfer_network_write_pipe() would be right place to check the result of write(). You ought to be able to immediately retry if you get EINTR (see glibc's TEMP_FAILURE_RETRY)

sim642: This isn't checking the result of the write() but the result of the poll() above. `SIGWINCH` caused that poll() to return `EINTR` making it think the poll() failed, while it actually did not.

PR #684, opened 2016-02-06
irc: add separate nullable options for kick/kill message coloring (closes #683)

### LINKED ISSUE #683: Add an option for changing kick message color
Similar to irc.color.reason_quit

sim642:
Add new options `irc.color.message_kick` and `irc.color.reason_kick` to be used for coloring KICK and KILL messages. These options have `irc.color.message_quit` and `irc.color.reason_quit` as parent options respectively not to change any users' current look.


Zarthus:
Nice! I like having my quit messages green whilst the kick ones are red.

Looking at the changes it looks like the option is merged, meaning you can't configure the color of kill messages. I may be wrong here - feel free to correct me if I am, but the following paragraphs explain why I'd like them to be separate (or at least kill not included in the list):

Is it possible to keep KILL's green (they are often less warranted: like in the case of NickServ REGAIN and do not need my attention) whereas KICKs are a channel level event (and especially in channels I op) would like to notice it as soon as it happens.

I feel by adding kills to this it has some sort of "diminishing return" as far as my need-for-attention goes, as the only kills worth looking at are those issued by an existing person, and not a service. Now I personally filter them out already, so it's not a biggie for me. But I would still like to point out the difference between a kick being mostly a human-issued thing, and a kill being often automated.

What's your mind on this?

sim642:
Currently I decided to split the options between self-caused (PART, QUIT) and other users' caused (KICK, KILL). To me it made more sense to have KILL with KICK than QUIT due to what it represents (a network-wide KICK).

If KILL had even separate color options than KICK, then it could even more so be argued that QUIT should have separate color options from PART as well. So we'd end up with 4 (PART/QUIT/KICK/KILL) * (message/reason) = 8 color options instead of the 2 which existed previously, which I think would be somewhat large degree of customization that would barely get used to its full extent. The total of 2 * 2 = 4 color options after this PR seems like a reasonable amount to me.

It'd still be possible to produce different coloring for KICK and KILL using a `hook_modifier` trigger to replace the color code.

Zarthus:
What [mIRC](http://i.imgur.com/Fw2SGQo.png) does (and that's where I get my inspiration from) is allow customization for every event there is, rather than by grouping them for what they are. I had personally imagined weechat has a similar thing going for msgbuffer.* as eventcolor.*, but that's a mistake on my part, and not in the scope of this Pull Request.

With the smart trigger and filters being a thing this really is not much of a concern to me, I'd just personally feel grouping KILL with KICK is a mistake. While logically it makes sense, in practice they both produce a while different kind of content - one which is relevant, the other is not.

I see where you're coming from with network-wide-kick, I just think KILLs generate too much (redundant) noise to the contrary of the far more useful KICK. Being used to mIRC grouping colours by event, at its core a KILL is still a QUIT, and a KICK is an entirely new event. All in all, I'd personally prefer that the colour KILLs represent the same colour as a QUIT message. 

Regardless on how this pull request ends up being merged, the ability is indeed there to still make it look like however I'd desire with use of triggers. I do however think that it's worth asking what other people might prefer to see which is the better option.

sim642:
@Zarthus I was told the following in #weechat:
```
< sim642> I still haven't heard any additional feedback on https://github.com/weechat/weechat/pull/684
< sim642> The question is: do you think KILL messages should have the same colors as QUIT or the same as KICK?
< grawity> given that you only receive a KILL if you're the victim, I'd say neither
< grawity> make them bright red and maybe blinking
< sim642> if someone else gets killed it shows that as a QUIT?
< grawity> yes
< sim642> in that case coloring it the same as KICK wouldn't hurt at all
```

It means that the coloring issue with nick regain `KILL`s won't be a problem as only that person gets the `KILL` while everyone else sees it as a `QUIT` with a special message.

Zarthus:
I was under the impression it's a KILL event too when you are killed yourself. But instead its an ERROR, so that's okay.

If other people's kills don't display differently, I'm fine with that.

flashcode:
Merged.
I removed the parent options to be consistent with all other colors in WeeChat (the new options have just same default color as the quit options).

PR #698, opened 2016-02-29
Add detection for Ruby 2.3

zeha:


flashcode:
Thanks for the PR.
@zeha: where is the file "ruby.pc" you mentioned in the debian bug #816312 ? I can't find it in any package in debian sid.

zeha:
@flashcode it's in ruby-dev/experimental. I'm not sure it will appear in sid before 1:2.3.0. Note that this is likely Debian specific (no idea what RH, Arch, etc. do here).

PR #700, opened 2016-03-05
gui: include short name checks when searching buffer by partial name

sim642:
When having two channels on different servers named #foo and ##foo, fixes a bug with "/buffer #foo" potentially switching to ##foo, while #foo would be a more accurate and intuitive match.

Also makes partial name match type priorities easier and less error prone to modify.


wolfsage:
Please merge this! (or something like it!)

My problem:

````
  /buffer d<tab> completes to '/buffer dom' which is a private message to dom.
````
However if I hit enter, I'm taken to the channel '#random'.

Exact-match first would be great

flashcode:
Rebased onto main, then merged.

Added:

- a commit to fix match order: end before middle (see d1a5796949c26d4f528f3520bc27c588cbee290d)
- unit tests to cover match order

PR #708, opened 2016-03-28
Update channel set bitlbee command in FAQ

migueldvb:
Update channel set bitlbee command to include the channel name.


flashcode:
I think it will be confusing for users to have such command as examples: many users just type the command without changing it (and then replacing the channel id).
If the current example is not enough, maybe another sentence should be added below the command to explain that an optional channel id is allowed.

migueldvb:
Without the channel name it looks like the command does not work so this gives a clue to the user what they need. I can add a sentence about the command in the English FAQ only.

arza-zara:
Channel id is optional, and unnecessary for FAQ in my opinion.

flashcode:
Yes the command works for me without any channel id.
So I think adding that would confuse users and is not needed in the FAQ, I agree with @arza-zara.

PR #713, opened 2016-04-08
fifo: Open the fifo for writing too

sardemff7:
That will make it survive, avoiding the need to reopen it.

### DIFF
--- a/src/plugins/fifo/fifo.c
+++ b/src/plugins/fifo/fifo.c
@@ -99,9 +99,9 @@ fifo_create ()
     /* create FIFO pipe, writable for user only */
     if (mkfifo (fifo_filename, 0600) == 0)
     {
-        /* open FIFO pipe in read-only, non-blocking mode */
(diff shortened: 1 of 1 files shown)


flashcode:
Hi @sardemff7,

Sorry for the very long time before reviewing this PR.

Could you please explain me the exact use case fixed by this PR? If possible with steps to reproduce the issue, so I can check if the PR correctly fixes it.

Thanks.

sardemff7:
It fixes a race condition when writing to the fifo via e.g. a shell script.
If you run multiple `echo /command > fifo` from a script, there is a chance for any non-first write to happen while the plugin is re-opening the fifo. That can lead to either ignored command (if the fifo is not removed yet but past the read error) or worse, the creation of a file (by the shell) preventing Weechat to re-create the fifo.
It can happen if several scripts are run in parallel too.

Opening it for both reading and writing, it will never die since there will always be at least one writer.

PR #716, opened 2016-04-15
core: add Portuguese translation

vascool:


flashcode:
Thank you for these translations, great work!
For you info: I fixed some minor whitespace (at the end of translations), and there are 3 untranslated messages.

PR #731, opened 2016-05-08
core: fix the hardcoded separator in string_split_command

irth:
The separator for string_split_command was hardcoded (`;`) instead of using the one from the function's arguments.
Not only did it not split strings with a different separator, it crashed WeeChat when `;` existed somewhere in the string.
It probably went unnoticed because in WeeChat's codebase it is only used with `;` as a separator.
I have modified the unit test to check whether it works with separator different than `;`.

### DIFF
--- a/src/core/wee-string.c
+++ b/src/core/wee-string.c
@@ -2118,7 +2118,7 @@ string_split_command (const char *command, char separator)
     while(*ptr != '\0')
     {
         type = 0;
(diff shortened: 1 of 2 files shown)


flashcode:
Thanks for the double fix: the function itself (bug older than 10 years old: a62a002522c5341a483e7058aba761001377b439), and the tests were bad: I wonder why I didn't check with different separators.

PR #732, opened 2016-05-08
gui: allow bar background color configuration in inactive windows

sim642:
This patch adds the (non-default) option to change background color (also modifiable via option) for bars in inactive windows (if enabled by an existing option). If enabled, it does not affect root bars or bars with `default` background color (to keep the look of input bar). Foreground colors remain as before.

_Feature requested by passbe in #weechat:_

```
< passbe> Is there anyway to switch the color of a bar when the window the bar is shown
          in is inactive? Documentation doesn't show anything. Would this even be
          possible as what would happen when you have 2 windows showing the same buffer...
```

### DIFF
--- a/src/core/wee-config.c
+++ b/src/core/wee-config.c
@@ -97,6 +97,7 @@ struct t_config_option *config_look_buffer_search_regex;
 struct t_config_option *config_look_buffer_search_where;
 struct t_config_option *config_look_buffer_time_format;
 struct t_config_option *config_look_color_basic_force_bold;
+struct t_config_option *config_look_color_inactive_bar;
 struct t_config_option *config_look_color_inactive_buffer;
 struct t_config_option *config_look_color_inactive_message;
 struct t_config_option *config_look_color_inactive_prefix;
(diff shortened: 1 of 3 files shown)


flashcode:
I'm currently reimplementing this as a new bar option "color_bg_inactive", so that you can customize the behavior per bar, and the global option to enable inactive bg color in bar is not needed any more.

To keep scripting API compatible, this new color will not be set directly in scripting API using function "bar_new", but it can be set once the bar is created.
But the C API will be updated and require a new argument "color_bg_inactive" after the argument "color_bg".

flashcode:
By the way, I plan to use inactive background color "darkgray" for default bars status and title.

flashcode:
I'll better update scripting API as well and fix existing scripts, which is better and consistent with the C API. It just requires some work to update the **14** affected scipts (sorted alphabetically):

- buddylist.pl
- buffers.pl
- chanmon.pl
- chanop.py
- highmon.pl
- iset.pl
- menu.pl
- moc_control.py
- newsbar.pl
- pv_info.pl
- rssagg.pl
- urlbar.py
- urlselect.lua
- vimode.py

weechatter:
some scripts are obsolete, like buffers.pl and iset.pl

flashcode:
@weechatter: you're right, I'll remove them soon.

PR #734, opened 2016-05-13
irc: notify: allow adding/removing multiple nicks

### LINKED ISSUE #689: notify: allow adding/removing multiple nicks
```
[irc]  /notify  add <nick> [<server> [-away]]
                del <nick>|-all [<server>]
```

Currently it seems that only one nick can be added or deleted from notify at the time unless this help text is outdated.

I would like to add/del multiple nicks and it would be a lot easier with typing just one line instead of separate for them all.

I don't know about ISON, but MONITOR allows multiple nicks to be added at the same time. If it's not possible with ISON, WeeChat could still accept all nicks at the same time and then send invidual ISON queries to them.

* http://ircv3.net/specs/core/monitor-3.2.html#monitor--targettarget2

* * * * *

* WeeChat 1.5-dev (git: v1.4-25-gcd23a5f) Copyright (C) 2003-2016, compiled on Jan 25 2016 13:28:38
* Debian GNU/Linux 7.9 (wheezy)

svkampen:
This patch adds support for adding/removing multiple nicks to/from the notify list, fixing issue #689.
One can now use `/notify add|del foo,bar` to add or delete notifications for foo and bar.

Example:
`/notify add foo,bar`

Output:

```
irc: notification added for foo
irc: notification added for bar
notify: foo is offline
notify: bar is offline
```

### DIFF
--- a/src/plugins/irc/irc-command.c
+++ b/src/plugins/irc/irc-command.c
@@ -3462,7 +3462,8 @@ IRC_COMMAND_CALLBACK(notice)
 IRC_COMMAND_CALLBACK(notify)
 {
     struct t_irc_notify *ptr_notify;
-    int i, check_away;
+    int i, check_away, num_nicks;
+    char **nicks;
 
     IRC_BUFFER_GET_SERVER(buffer);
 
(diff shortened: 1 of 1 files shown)


src/plugins/irc/irc-command.c
```diff
@@ -6526,7 +6526,9 @@ irc_command_init ()
            "  notify when \"toto\" joins/quits freenode server:\n"
            "    /notify add toto freenode\n"
            "  notify when \"toto\" is away or back on freenode server:\n"
-           "    /notify add toto freenode -away"),
+           "    /notify add toto freenode -away",
+		   "  notify when \"toto\" or \"gary\" join/quit current server:\n",
```
sim642: Bad indentation

src/plugins/irc/irc-command.c
```diff
@@ -6526,8 +6526,8 @@ irc_command_init ()
            "  notify when \"toto\" joins/quits freenode server:\n"
            "    /notify add toto freenode\n"
            "  notify when \"toto\" is away or back on freenode server:\n"
-           "    /notify add toto freenode -away",
-		   "  notify when \"toto\" or \"gary\" join/quit current server:\n",
+           "    /notify add toto freenode -away\n"
```
sim642: This should probably be squashed into the original commit.

src/plugins/irc/irc-command.c
```diff
@@ -3518,47 +3519,53 @@ IRC_COMMAND_CALLBACK(notify)
...
-                _("%s%s: error adding notification"),
-                weechat_prefix ("error"), IRC_PLUGIN_NAME);
+            if ((ptr_server->monitor > 0)
+                && (ptr_server->notify_count >= ptr_server->monitor))
+            {
+                weechat_printf (
+                    ptr_server->buffer,
+                    _("%sMonitor list is full (%d)"),
+                    weechat_prefix ("error"), ptr_server->monitor);
+                continue;
```
sim642: If monitor list is full, there is no point in showing the same error for any of the nicks remaining to be processed.

PR #738, opened 2016-05-19
irc: fix NULL pointer dereference in 734 command callback

scumjr:
The following server command makes weechat crash: ``:blah 734 a b c\r\n``. This patch (which is copy/paste of ``IRC_PROTOCOL_CALLBACK(733)``) should fix this issue.


flashcode:
Thanks for the fix.

PR #739, opened 2016-05-20
doc: OSX gnutls path for Freenode

recurrence:
The default Freenode SSL fix does not work in OSX.  However, when the homebrew openssl package is installed (common in osx environments), a pem file containing the necessary certs is available.


recurrence:
good to merge?

PR #756, opened 2016-07-03
clarification of /secure help doc

wilmo:


weechatter:
In buffer title of secure-buffer you can read: Keys: [alt-v] Toggle values

On the other hand key bindings are described in /help script

flashcode:
Hi,
Thanks for your pull request.
The files in `doc/*/autogen/*` must not be modified by hand, they are automatically built by a script, and then included in the docs (see the header in these files).
The file to change was src/core/wee-command.c, which defines core commands, with the associated help text (this text is then translated with gettext and used to build autogen files).

Issue #754, opened 2016-06-24
Date is not translated properly when LC_TIME is in a different language


allan2:
Using `%a, %d %b %Y %T` does not properly use the locale format of LC_TIME, resulting in a partial loss in translation.

Example:
`LC_TIME=en_CA.UTF-8 weechat`

> Channel created on Fri, 24 Jun 2016 13:33:08

`LC_TIME=zh_CN.UTF-8 weechat`

> Channel created on 五, 24 6月 2015 13:33:08

So the date is using something like `%a, %d %b %Y %T`

Note:

`LC_TIME=en_CA.UTF-8 weechat`

> Fri Jun 24 13:33:08 EDT 2016

The date elements are almost identical in English (there is no comma after `%a`).

`LC_TIME=zh_CN.UTF-8 weechat`

> 2016年 06月 24日 星期五 13:33:08 EDT

Although the 月 is hardcoded into `%b`, the other essential elements are lost in translation.

Using WeeChat 1.5 on Gentoo

sim642:
That is how locales work: they specify the content of the specific date elements but don't modify the formatting string. The date elements are part of the locale, the formatting string is part of the code which formats the date, i.e. WeeChat.

allan2:
I am aware.

The issue is in the current WeeChat formatting string, `%a, %d %b %Y %T`, works for English but not for other `LANG` or more specifically, `LC_TIME`.

It should take into account the locale by drawing from `locale date_fmt`, i.e. `date +%c`.

Using just `%c` would essentially solve the problem. It would introduce time zone to the date format, but that is trivial to remove.


Posted so you can see the format string differences:

`LC_TIME=zh_CN.UTF-8 locale date_fmt`

> %Y年 %m月 %d日 %A %H:%M:%S %Z

`LC_TIME=en_CA.UTF-8 locale date_fmt`

> %a %b %e %H:%M:%S %Z %Y

sim642:
> It would introduce time zone to the date format, but that is trivial to remove.

Not really.

If you're thinking of removing the last space-separated token from the output with `%c` then we're bound to have another issue where someone comes along and says that in their locale date format string for `%c` does not contain the timezone as the last thing (your `en_CA` example). The whole reason why all the %-codes exist, to allow for fine-tuned control.

allan2:
[...]

PR #766, opened 2016-07-28
locale-sensitive dates (ONLY IF IT HAS TZ) #754

allan2:
in regards to #754, replace full date strings that have timezone with `%c` which has timezone

date strings without timezone remain locale-insensitive


flashcode:
Changing the CTCP time format to `%c` will not have the expected result, because WeeChat is setting the locale to "C" (which means no locale) before building the string to return, see: https://github.com/weechat/weechat/blob/master/src/plugins/irc/irc-ctcp.c#L453

It is set to "C" because the nick asking the CTCP TIME will always receive English text (the user may not speak the same language as the sender).

About the `%H:%M:%S` converted to `%T`, what is the difference?

(cc #754)

PR #771, opened 2016-08-06
Use res_init before getaddrinfo.

ronin13:
With weechat, if the network changes in anyway that causes resolv.conf
to be updated, that is not picked up, resulting in

    irc: connecting to server XXXX.com/1026 (SSL)...
    irc: address "XXX.com" not found
    irc: error: Name or service not known
    irc: reconnecting to server in 30 seconds

This happens, especially when one connects to an irc server behind a
VPN.

This can also happen when there is no network connectivity weechat was started
but later network is connected to.

The fix is simple, it relies on calling  res_init (3) before getaddrinfo in
network_connect_child and other places.

res_init reloads the resolv.conf and getaddrinfo does not fail with
address resolution error.

### DIFF
--- a/src/core/wee-network.c
+++ b/src/core/wee-network.c
@@ -43,6 +43,7 @@
 #include <netinet/in.h>
 #include <arpa/inet.h>
 #include <netdb.h>
+#include <resolv.h>
 #include <errno.h>
 #include <gcrypt.h>
 #include <sys/time.h>
@@ -310,6 +311,9 @@ network_resolve (const char *hostname, char *ip, int *version)
         *version = 0;
 
     res = NULL;
+
+    if (res_init() != 0)
+	return 0;
 
     if (getaddrinfo (hostname, NULL, NULL, &res) != 0)
         return 0;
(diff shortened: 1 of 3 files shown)


flashcode:
Test are failing, I think some `#include` are missing?

flashcode:
And what about the getaddrinfo in src/plugins/irc/ir-server.c ?

ronin13:
Thanks, I have fixed them and pushed.

src/core/wee-network.c
```diff
@@ -311,6 +312,9 @@ network_resolve (const char *hostname, char *ip, int *version)
 
     res = NULL;
 
+    if (!res_init())
```
flashcode: According to `man res_init`, the function returns 0 on success, -1 if an error occurs.
So I think the test should be: `if (res_init() != 0)`
Same problem in other tests in this commit.

ronin13:
Sorry about that, I have fixed it now.  

The reason I missed it was there was no error handling (when I tested myself) in network_connect_child (which is mainly used).  I was not sure how to handle error in that path there. I have added it now.

flashcode:
Thanks, merged.
Finally, I removed the check on `res_init()` return code: if this call fails, the connection can still be OK. And in worst case, you'll end with the error `Name or service not known`.

ronin13:
Yeah, should be fine skipping that. :+1:  

Though I was wondering if res_init() is failing,  something must be terribly wrong :), so changes of getaddr* working later are slim. 

Thanks! (This bug was causing me pain for quite some time (since I vpn a lot), didn't have chance to look at it all this time).

PR #772, opened 2016-08-07
core, gui, plugins: Fix resource leaks and possible NULL dereferences.

ronin13:
Dangling file pointer fixed in util_file_get_content.
Resource leak fixed in xfer_chat_open_buffer.
Some minor NULL deference fixes.

.gitignore also updated with cmake artifacts.

=========
Found them when I was testing infer (https://github.com/facebook/infer) with weechat.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,10 +1,12 @@
 # ignored files for Git
 
 *.a
+*.cmake
 *.gmo
 *.la
 *.lai
 *.lo
+*.out
 *.m4
 *.o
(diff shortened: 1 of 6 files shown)


flashcode:
Thanks for these fixes, I merged them with some minor changes.

For the .gitignore file, I committed only the symbolic link on binary (weechat-curses).
The other changes are not needed if you build in a build directory, which I recommend (there's a line with `build*/*`).

PR #778, opened 2016-08-12
plugin: Avoid the conflict with gnutls and guile memory manager.

ronin13:
This fix avoids the crash detailed in https://github.com/weechat/weechat/issues/779 with
gnutls when weechat exits.

Workaround (without this fix) is to build weechat without guile plugin.


flashcode:
Waiting for a reply in the issue #779.

flashcode:
Issue #779 was not reproduced, so I close this PR.

PR #785, opened 2016-08-25
core: Allow for multiple gnutls CA files (closes #569)

### LINKED ISSUE #569: Allow specifying multiple root CA files for GnuTLS
Right now to trust a root CA of some IRC server with custom certificate there are two options:
1. Append that certificate into system wide CA certs file (might not be accessible)
2. Create a new certs file which is a concatenation of the system's and your own (not easily kept in sync with the system's one)

This is why there should be a way to simply specify multiple files which contain CA certs to trust.

fceschmidt:
This has been tested with two CA files, the merged one from my default SSL certificate store and the Hackint certificate, which is also my use case for this feature.

For specifying multiple CA files in weechat.conf, separate CA files with a colon in between. They will be added to the trusted certificates at start-up time (upon the call to `network_set_gnutls_ca_file`). I additionally amended `string_expand_home` in such a fashion that it is now able to replace `~/` with `/home/...` not only at the beginning of the string, but also after each colon in the string.

I propose to add some unit tests for the additional functionality of `string_expand_home` in a separate commit.

### DIFF
--- a/src/core/wee-network.c
+++ b/src/core/wee-network.c
@@ -107,9 +107,32 @@ network_set_gnutls_ca_file ()
         ca_path2 = string_replace (ca_path, "%h", weechat_home);
         if (ca_path2)
         {
-            gnutls_certificate_set_x509_trust_file (gnutls_xcred, ca_path2,
-                                                    GNUTLS_X509_FMT_PEM);
+            char *ca_path3 = ca_path2;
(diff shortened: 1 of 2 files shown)


fceschmidt:
Apparently I have some errors left in my code, I'll go back to testing first.

Issue #658, opened 2016-01-18
TLS cert validation requires intermediate CAs in ca_file


azrdev:
I try to connect to hackint and freenode using TLS/SSL and verifying the servers' certificate. To this avail, I downloaded the root CAs for each of the networks, and `cat`ed them (in PEM format with LF line endings) to `~/.weechat/ssl/CAs.pem`. I set the weechat option `weechat.network.gnutls_ca_file` to `%h/ssl/CAs.pem`.

With this setup, I expect to be able to connect to the networks. However, I would always get an error about the server certificate not being trustworthy. I can fix this by extracting the intermediate certificate from the server [1] and appending it to my CA file, but this is not how things should work.

Weechat 1.4
GnuTLS 3.3.20

[1]: Using `openssl s_client -connect server:6697 -verify_return_error -showcerts | less` - the certificates are in `BEGIN CERTIFICATE … END CERTIFICATE` blocks, they send two: first the servers' own cert, after that the intermediate. The root cert is not sent.

PR #786, opened 2016-08-25
core: Allow for multiple gnutls CA files (closes #569)

### LINKED ISSUE #569: Allow specifying multiple root CA files for GnuTLS

fceschmidt:
This has been tested with two CA files, the merged one from my default SSL certificate store and the Hackint certificate, which is also my use case for this feature.

For specifying multiple CA files in weechat.conf, separate CA files with a colon in between. They will be added to the trusted certificates at start-up time (upon the call to `network_set_gnutls_ca_file`). I additionally amended `string_expand_home` in such a fashion that it is now able to replace `~/` with `/home/...` not only at the beginning of the string, but also after each colon in the string.

I added some unit tests as well for the additional functionality of `string_expand_home`.

P.S.: Sorry for creating two pull requests, I should really embrace TDD...


stfnm:
For splitting the strings at the colon you might also wanna consider using weechat's [string_split()](https://github.com/weechat/weechat/blob/master/src/core/wee-string.c#L1783), which could simplify the code a little further.

fceschmidt:
Thanks for the hint, I did that and also fixed one more thing about Windows compatibility by adding a platform-dependent macro definition in `weechat.h`.

sim642:
In my opinion `string_expand_home` function should not be dealing with separators in the case of multiple delimited paths. Wouldn't it simply be possible to achieve the same result with a lot less effort by calling `string_expand_home` on every one of the split parts?

src/core/wee-network.c
```diff
@@ -107,9 +107,21 @@ network_set_gnutls_ca_file ()
...
+            char **single_paths;
+
+            single_paths = string_split(ca_path2, PATH_SEPARATOR, 0, 0, 
+                                        &num_paths);
+
+            for (i = 0; i < num_paths; i++)
+            {
+                gnutls_certificate_set_x509_trust_file (gnutls_xcred,
+                            single_paths[i], GNUTLS_X509_FMT_PEM);
+                free (single_paths[i]);
```
sim642: Unnecessary, use `string_free_split` instead.

src/core/wee-network.c
```diff
@@ -107,9 +107,21 @@ network_set_gnutls_ca_file ()
...
+                single_paths[i] = NULL;
+            }
             free (ca_path2);
+            free (single_paths);
```
sim642: Use `string_free_split` here.

src/core/wee-network.c
```diff
@@ -107,9 +107,21 @@ network_set_gnutls_ca_file ()
         ca_path2 = string_replace (ca_path, "%h", weechat_home);
         if (ca_path2)
         {
-            gnutls_certificate_set_x509_trust_file (gnutls_xcred, ca_path2,
-                                                    GNUTLS_X509_FMT_PEM);
+            int num_paths, i;
```
sim642: It is a convention throughout WeeChat source to declare variables at the very beginning of a function.

fceschmidt:
Just adding that this also provides a solution for #658

src/core/wee-network.c
```diff
@@ -96,23 +96,30 @@ void
...
+    char *ca_path, *ca_path2, **single_paths;
+    int i;
 
     if (weechat_no_gnutls)
         return;
 
-    ca_path = string_expand_home (CONFIG_STRING(config_network_gnutls_ca_file));
-    if (ca_path)
+    single_paths = string_split(CONFIG_STRING(config_network_gnutls_ca_file), 
+                                PATH_SEPARATOR, 0, 0, NULL);
```
nimaje: I don't think it is a good idea to have the seperator OS dependent, if it doesn't have to be like here

flashcode:
This option has been replaced by `weechat.network.gnutls_ca_user` in WeeChat ≥ 3.2 and it supports several paths separated by `:`.

So I close this PR.

Issue #573, opened 2015-11-03
irc: not enough memory


kashike:
I'm encountering the following issue:

```
08:28:10 =!= | irc: reading data on socket: error -54 Error in the pull function.
08:28:10  -- | irc: disconnecting from server...
08:28:10  -- | irc: disconnected from server
08:28:10  -- | irc: reconnecting to server in 10 seconds
08:28:20  -- | irc: reconnecting to server...
08:28:20  -- | irc: connecting to server ... (SSL)...
08:28:20 =!= | irc: not enough memory
```

on reconnect, the same thing (`irc: not enough memory`) happens. The only solution is to restart weechat. So far, from what I recall, it is only occurring on one network (rizon).

`OpenSSL 1.0.1f 6 Jan 2014`

```
10:59:08     | Scripts loaded:
10:59:08     |  * ia   buffers.pl 5.1 (Sidebar with list of buffers)
10:59:08     |    ia   highmon.pl 2.5 (Highlight Monitor)
10:59:08     |  * ia   iset.pl 3.9 (Interactive Set for configuration options)
10:59:08     |    ia   autojoin.py 0.2.5 (Configure autojoin for all servers according to currently joined channels)
10:59:08     |    ia   autosort.py 2.5 (Automatically or manually keep your buffers sorted and grouped by server.)
10:59:08     |    ia   colorize_nicks.py 21 (Use the weechat nick colors in the chat area)
10:59:08     |    ia   prism.py 0.2.8 (Taste the rainbow.)
10:59:08     |  * ia   urlserver.py 1.9 (Shorten URLs with own HTTP server)
10:59:08     |    ia   weestats.py 0.2 (Useless statistics about your open buffers and windows)
```

```
WeeChat 1.4-dev [compiled on Sep 17 2015 06:10:06]
Upgraded 1 time, first start: Wed, 21 Oct 2015 23:07:37
```

```
Message filtering enabled
No message filter defined
```

```
Triggers enabled
List of triggers:
  beep: print =? /1
  cmd_pass: modifier(5000|input_text_display;5000|history_add;5000|irc_command_auth) ~1
  msg_auth: modifier(5000|irc_message_auth) ~1
  server_pass: modifier(5000|input_text_display;5000|history_add) ~1
```

flashcode:
Hi,
What is the OS used (and its version)?
And how much memory WeeChat is using? (RES column in command `top`)
You probably reached the max memory available, which causes such message in WeeChat.

[...]

Issue #358, opened 2015-03-10
Failed SSL connections leak sockets


WGH-:
Apparently, every time SSL connection is failed, a socket is leaked. My weechat has been running for a week I think, trying to connect to some server and always failing it due to certificate problem, and open socket count just keeps increasing.
```
% sockstat | grep weechat | wc -l
569
```
According to `netstat`, they're all in the closed state.

I'm running weechat on FreeBSD, gnutls is used.

flashcode:
Hi,
What's your WeeChat version?
And when you say "SSL connection fails", for which reason? Do you have examples of errors displayed in server buffer?

WGH-:
Version 1.1.1

```
irc: reconnecting to server in 30 minutes
irc: reconnecting to server...
irc: connecting to server irc6.rizon.net/9999 (SSL)...
gnutls: connected using 2048-bit Diffie-Hellman shared secret exchange
gnutls: receiving 1 certificate
 - certificate[1] info:
   - subject `C=RZ,ST=Nowhere,L=Rizon,CN=irc.rizon.net,O=Rizon IRC Network,OU=irc.broke-it.com', issuer `C=RZ,ST=Nowhere,L=Rizon,CN=irc.rizon.net,O=Rizon IRC Network,OU=irc.broke-it.com', RSA key 2048 bits, signed using RSA-SHA1, activated `2013-07-24 21:52:39 UTC', expires `2016-07-23 21:52:39 UTC', SHA-1 fingerprint `150159fe4c0916c13424a312bc2832232922d9e1'
gnutls: the hostname in the certificate does NOT match "irc6.rizon.net"
gnutls: peer's certificate is NOT trusted
gnutls: peer's certificate issuer is unknown
irc: TLS handshake failed
irc: error: Error in the certificate.
irc: reconnecting to server in 30 minutes
```

In case it's helpful, this's how I connect to that server:
```
rizon.addresses = "irc6.rizon.net/9999"
rizon.proxy
rizon.ipv6
rizon.ssl = on
rizon.ssl_cert
rizon.ssl_priorities
rizon.ssl_dhkey_size
rizon.ssl_fingerprint = ""
rizon.ssl_verify
rizon.password
rizon.capabilities
rizon.sasl_mechanism
rizon.sasl_username
rizon.sasl_password
rizon.sasl_timeout
rizon.sasl_fail
rizon.autoconnect = on
rizon.autoreconnect
rizon.autoreconnect_delay
rizon.nicks
[...]

PR #801, opened 2016-09-19
Fix for unclosed/leaking sockets from issue #358

kg6hum:
The issue was server->sock not being set unless the connection was successful. In the case of SSL connections, a connection could be established but not be successful because it wasn't an SSL port or the SSL certificate did not validate.  In those cases irc_server_close_connection() was called, but it was checking for server->sock != -1 before calling gnutls_bye() or close().  Since close() was never called for the socket, it would get stuck in the CLOSE_WAIT state until weechat was closed.

This will also fix issue #573 when the unclosed sockets exceed the fd limit.


Zarthus:
I hope @flashcode is willing to merge this before 1.6's release, it's been affecting a lot of people.

Thanks for figuring out what was causing it!

PR #814, opened 2016-10-02
core: fix build with netbsd-curses [0]

rofl0r:
eat_newline_glitch is read-only on netbsd curses, so doing an assignment
breaks the build.
fix it by checking if we're using ncurses, which is obviously what the code's
author targeted.

[0] https://github.com/sabotage-linux/netbsd-curses

### DIFF
--- a/src/gui/curses/gui-curses-term.c
+++ b/src/gui/curses/gui-curses-term.c
@@ -45,7 +45,7 @@
 void
 gui_term_set_eat_newline_glitch (int value)
 {
-#ifdef HAVE_EAT_NEWLINE_GLITCH
+#if defined(NCURSES_VERSION) && defined(HAVE_EAT_NEWLINE_GLITCH)
     eat_newline_glitch = value;
 #else
     /* make C compiler happy */


rofl0r:
does the "waiting info" label you added indicate that you're waiting for more, yet unspecified info from me, @flashcode ?

flashcode:
@rofl0r : yes, see my question above in review, I think your changes could introduce a regression.

rofl0r:
i'm not seeing any reply from you (maybe a browser problem?). would you be so kind as to repeat your reply in a normal comment ? thanks!

flashcode:
@rofl0r: I asked:

> Are you sure this can not introduce a regression?
> On ncurses versions where NCURSES_VERSION is not defined (I don't know if it's possible, but who knows...)
> Isn't there another way to check if we are using netbsd-curses? Something specific to this version, so we are sure this will not break any other ncurses version.

rofl0r:
thanks for the repost.
all ncurses versions since at least 4.1 (19 years old) define NCURSES_VERSION so it works everywhere.
for reference: http://ncurses.scripts.mit.edu/?p=ncurses.git;a=blob;f=include/curses.h.in;h=43e474ba449b8af36170d1dfec37f838d4444791;hb=3a9b6a3bf0269231bef7de74757a910dedd04e0c#l35 (commit id is 4.1 tag)

rofl0r:
ok i can see that you don't want to hardcode assumptions about library versions into the code.
would you accept a configure compile test that checks if eat_newline_glitch is writable and then use something like `#ifdef HAVE_WRITEABLE_EAT_NEWLINE_GLITCH` instead ?

flashcode:
@rofl0r : yes that would be better to add that before compilation then. If you do that, don't forget you must update both CMake and configure scripts (CMake is preferred way to compile weechat, but configure is still there for historical reasons and should work too).

rofl0r:
turned out that the issue is due to a bug in configure ^.
it already checked if eat_newline_glitch can be written to, but due to a typo in the variable name it didnt actually process the result of the configure check.

PR #820, opened 2016-10-12
Add server option umodes (closes #377)

### LINKED ISSUE #377: Add irc.server.xxx.umodes
I know you can do this with `irc.server.xxx.command` already, but I think adding a more explicit option could be nice.

sim642:


sim642:
Things to think/discuss/decide about:

1. Should `usermodes` be the option name instead?  
    Currently WeeChat shows the setting of umodes in server buffer as `-- User mode [+irwxz] by sim642`.
2. Should the option's helptext be more descriptive?

arza-zara:
1. I suggest singular `usermode`.
2. The helptext is sufficient.

PR #826, opened 2016-10-20
core: handle lists in hook_process_hashtable for urls (closes #219)

### LINKED ISSUE #219: Add possibility to set headers in hook_process_hashtable
Hi,

It would be useful to be able to set custom headers in hook_process_hashtable, like the Content-Type as discussed on irc. guile-curl takes an alist for example, with a key and a value for each.

soolar:
Adds support for every curl option that uses a curl_slist, not just httpheader, but I didn't see issues opened for any of the other options this enables.


flashcode:
Merged, thanks.

PR #834, opened 2016-11-01
lua: Use lua_pushinteger for integer values in Lua 5.3+

tomoe-mami:
In Lua 5.3+, when displaying values from Weechat functions that return integer the output will have `.0` appended. This is because `lua_pushnumber` since 5.3 only pushes float value and float to string conversion now keeps the fractional part.

To reproduce, build Weechat with Lua 5.3+ and load [this simple script](http://sprunge.us/bVNE). It should print the current buffer number followed by `.0`.

This PR replaces `lua_pushnumber` with `lua_pushinteger` to fix this issue.

### DIFF
--- a/src/plugins/lua/weechat-lua-api.c
+++ b/src/plugins/lua/weechat-lua-api.c
@@ -82,13 +82,21 @@
     if (__string)                                                       \
         free (__string);                                                \
     return 1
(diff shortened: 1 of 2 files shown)


flashcode:
Thanks!

PR #866, opened 2016-12-15
Simplify memory management

DemiMarie:
Many pointers were being compared with NULL before being freed.
This is not required.  One function allocated a struct with
`malloc` and then zeroed out its fields manually; call `calloc`
instead.  Use a macro to automate the `free(x); x = NULL` pattern.

### DIFF
--- a/src/core/wee-config-file.c
+++ b/src/core/wee-config-file.c
@@ -184,7 +184,7 @@ config_file_new (struct t_weechat_plugin *plugin, const char *name,
         new_config_file->name = strdup (name);
         if (!new_config_file->name)
         {
-            free (new_config_file);
+            FREE(new_config_file);
             return NULL;
         }
         length = strlen (name) + 8 + 1;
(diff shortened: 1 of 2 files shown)


src/core/wee-config-file.c
```diff
@@ -1234,7 +1196,7 @@ config_file_option_reset (struct t_config_option *option, int run_callback)
         {
             if (option->value)
             {
-                free (option->value);
+                FREE(option->value);
                 option->value = NULL;
```
arza-zara: [R1165](https://github.com/weechat/weechat/pull/866/files#diff-3ba66fdfee8c62012bd7f1fe08767041R1165) and [R1200](https://github.com/weechat/weechat/pull/866/files#diff-3ba66fdfee8c62012bd7f1fe08767041R1200) also removable?

src/core/wee-config-file.c
```diff
@@ -1196,11 +1161,8 @@ config_file_option_reset (struct t_config_option *option, int run_callback)
                     || (strcmp ((char *)option->value,
                                 (char *)option->default_value) != 0))
                     rc = WEECHAT_CONFIG_OPTION_SET_OK_CHANGED;
-                if (option->value)
-                {
-                    free (option->value);
-                    option->value = NULL;
-                }
+                FREE(option->value);
+                option->value = NULL;
```
arza-zara: [R1165](https://github.com/weechat/weechat/pull/866/files#diff-3ba66fdfee8c62012bd7f1fe08767041R1165) and [R1200](https://github.com/weechat/weechat/pull/866/files#diff-3ba66fdfee8c62012bd7f1fe08767041R1200) also removable?

PR #868, opened 2016-12-20
Grammar correction

neutric:
Documentation is commonly used in singular form even if it refers to multiple documents.


flashcode:
Thanks.

PR #870, opened 2016-12-22
core: fix misnamed configure cache variables

rofl0r:
these 3 configure tests were bogus in that they didnt return a result:
```
checking for flock() support...
checking for execinfo.h and backtrace...
checking for eat_newline_glitch support...
```

looking at config.log reveals:
```
configure:24327: checking for eat_newline_glitch support
configure:24344: gcc -c -g -O2 -DHAVE_GNUTLS -D_FILE_OFFSET_BITS=64 ...
conftest.c: In function 'main':
conftest.c:134:2: error: assignment of read-only location '*(cur_term->flags ...
configure:24344: $? = 1
configure: failed program was:
...
configure:24351: result:
```

but due to the misnamed variables configure ended up enabling
eat_newline_glitch.

fixes #814


flashcode:
Thanks.

PR #872, opened 2016-12-27
reduce GUI_BUFFERS_MAX to 100 to prevent denial of service attacks

R41z:


sim642:
This PR is directly limiting WeeChat's capabilities by not allowing over 100 buffers, meaning you won't be able to join more channels than that. There are plenty of users who have well over 100 buffers.

What DoS are you trying to prevent?

flashcode:
Hi,

I don't agree with this PR which limits drastically the number of buffers you can open.
I know lot of people using more than 100 buffers, such limit is very low.

If you really want to limit the number of opened buffers, an option should be added instead of an hardcoded constant. So that the user can control the max number of buffers, depending on its own needs.

R41z:
ok, maybe I went too far. let me explain what could go wrong... there are botnets that reconnect with different nicks to irc and send you spam, not just once or twice, but hundreds of times, and since I run weechat on a server, it's there 24/7, the problem is if the amount of buffers exceed the amount of files a process can open, this can crash weechat, because weechat opens a file for logging for each buffer.

I'm not sure if you get what I tried to say.

nyuszika7h:
You can set user mode `+g` (or `+R` to block just unregistered users, which has worked for me so far) to block that spam.

R41z:
this does not work in all networks, sorry.

weechatter:
you can also use query_blocker.pl

R41z:
not if weechat got this fixed, stop suggesting stupid ideas

weechatter:
less stupid idea then crippling weechat to only 100 buffers!! and there is nothing to "fix" in weechat to prevent spam bots opening query buffers. query_blocker is a solution for server does not support user mode +g / +R

R41z:
well, it makes no sense to allow weechat buffers to exceed the amount of allowed open files per process, that's my point

sim642:
If you have so many buffers that it's a problem then maybe increase the
limits in your system. Or actually deal with the spam, not pretend to solve
the problem by creating unreasonable limits for other users.

R41z:
Fine, you don't want to merge, I respect that.

flashcode:
@R41z: WeeChat should not crash when running out of file descriptors. If this happen when starting to log a buffer, the logging should just be disabled on this buffer.
If you experience a crash in such case, please file a bug.

PR #874, opened 2016-12-28
gui: add options for scroll bar item prefix and suffix

sim642:
Adds options for configuring `scroll` bar item's format runtime to something other than what's specified by the language.

*Feature requested by earnestly in #weechat:*
```
<@nils_2> don't understand what you want archer121
< earnestly> Quick question, for the 'scroll' item, can the display be changed so instead of
             '-MORE(n)-' I can just display 'n'?
<@nils_2> no. its hard-coded.
< earnestly> Ah, okay thanks for confirming.  I quickly scanned the options for anything
             relevant but couldn't find anything and pondered if it was
```

### DIFF
--- a/src/core/wee-config.c
+++ b/src/core/wee-config.c
@@ -186,6 +186,8 @@ struct t_config_option *config_look_scroll_page_percent;
 struct t_config_option *config_look_search_text_not_found_alert;
 struct t_config_option *config_look_separator_horizontal;
 struct t_config_option *config_look_separator_vertical;
+struct t_config_option *config_look_status_more_prefix;
+struct t_config_option *config_look_status_more_suffix;
 struct t_config_option *config_look_tab_width;
 struct t_config_option *config_look_time_format;
 struct t_config_option *config_look_window_auto_zoom;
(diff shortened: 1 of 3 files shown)


src/core/wee-config.c
```diff
@@ -3402,6 +3421,22 @@ config_weechat_init_options ()
         &config_check_separator, NULL, NULL,
         &config_change_buffers, NULL, NULL,
         NULL, NULL, NULL);
+    config_look_status_more_prefix = config_file_new_option (
+        weechat_config_file, ptr_section,
+        "status_more_prefix", "string",
+        N_("text to display before count for buffer with new data (status bar)"),
+        NULL, 0, 0, _("-MORE("), NULL, 0,
```
sim642: I'm not sure if this `_("-MORE(")` as a default value is still proper if we want to allow it to be translatable but modifiable by the user if they wish to do so.

flashcode: For me it's a problem: the default value of an option should not depend on the user locale: any default config file on any system must be the same.

The translated strings should not be customizable themselves, instead we should offer another way to customize this bar item if needed : I see for example two solutions:

1. add another bar item with the more value, without label
2. add options for bar items, so the "scroll" bar item could have an option to disable the label.

And by the way, a script can easily add such bar item.

PR #876, opened 2016-12-31
irc: fix buffer switching on manual join for forwarded channels

sim642:
Previously using option values
```
	irc.look.buffer_open_before_join off
	irc.look.buffer_switch_join on
```
and manually joining a channel which gets forwarded (e.g. `#linux` → `##linux-overflow` on freenode) the channel buffer for `##linux-overflow` was not switched to even though the option says it should
have.

This patch copies manual join and noswitch information for channels which get forwarded.


flashcode:
Thanks!

Issue #877, opened 2017-01-01
/exec -o: Option to control command execution


xnrand:
When running external programs with `/exec -o`, any output lines starting with a `/` will be run as weechat commands. A way to disable or control that behavior would be useful.

One approach would be to additional /exec flags similar to `-o`, for example:
- `-cmd` to run each line of output as a command
- `-otext` to output each line of output as text into the current buffer, not running any commands

flashcode:
I changed the default behavior of option "-o" which is now to NOT execute commands in output (which is more secure).
I hope it's OK for everybody to have such default.
WIth the new option "-oc", commands are executed.

weechatter:
perfect!

PR #883, opened 2017-01-07
exec: update help descriptions for command /exec to keep compatibilit…

l:
…y between options

Related-Bug: #877


flashcode:
Thanks, merged!
I just fixed the word "option" to use the plural form in case of multiple options listed after.

PR #885, opened 2017-01-09
freenas recvmsg() workaround

walker0643:


flashcode:
Hi,
Is this change still needed today?
If yes, can you tell me more about the changes and which problem it solves?

walker0643:
Hello - To be honest I don't remember what the issue was. This was a patch I was running in 2016/2017 because it made weechat usable on FreeBSD and I was doing a lot of BSD work at the time. Maybe take a look at FreeBSD's recvmsg() and see if it's still adding the extra SCM_CREDS before the SCM_RIGHTS structure?? If that has been changed in the last 6-7 years then I would guess that this patch is no longer required. Sorry I couldn't be more help.

flashcode:
Thanks @walker0643.
I regularly build and test WeeChat on FreeBSD, and connection to servers works fine without your changes.
So I close this PR.

PR #893, opened 2017-01-22
core/gui: added possibility to resize full windows parents

raspbeguy:
This PR improves the `/window resize` function:
when specifying `v` (vertically) or `h` (horizontally) before the value, this will affect the size of the nearest window parent that can be resized in the given direction.

For instance:
```
/window resize h+10
/window resize v20
```
Of course, the existing usage of `/window resize` is respected.


flashcode:
Thanks, works fine for me!
Merged with some minor changes.

PR #905, opened 2017-02-08
/build should be included by the repository

aiya000:
This is the one of idea :smile:

I think /build should be included, because it is more natural when an user builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@
 ABOUT-NLS
 autom4te*
 build*/*
+!build/*
 compile
 config.guess
 config.h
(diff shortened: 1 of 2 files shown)


flashcode:
Hi,
The `build` directory is just an example of name in docs, users can use any directory name here.
And it's a temporary directory to build WeeChat, I don't see any reason to « unignore » this in .gitignore or even include that in the package.

zao:
One advantage of having an empty `build` directory in the repo is that it's slightly more obvious that the project is capable of an out-of-tree CMake build.

stfnm:
Git doesn't allow committing empty directories, so you would have to put at least some dummy file in it.

Anyway, I don't really see much point in that. With cmake the build directory can be pretty much any place you want; like directly in the source tree, or any subdirectory of the source tree (named "build" or whatever else you prefer), or even some completely different place like "/tmp/build_dir_for_weechat".

aiya000:
!  
I understood it :smile: (It is an example name)  
Sorry !

Issue #141, opened 2014-07-26
Add option to read passphrase from another program's stdout


seschwar:
`msmtp` and `mpop` have a `passwordeval` option.  You can specify a command to run from who's standard output `msmtp` or `mpop` will read the password.

I would like to see this in WeeChat as well, since it will allow for a couple of things:
- integration with standalone password managers like `pass` and `pwsafe`
- integration with password managers of desktop environments through query programs like `gnome-keyring-query` and `kwalletcli`
- reading the passphrase from a file encrypted with `gpg` and therefore leave the prompting to `gpg-agent`
- spawn graphical password prompts with `kdialog`, `x11-ssh-askpass`, `zentiy` and others

shioyama:
:+1: this would be great to have

incertia:
are people still interested in this feature? i can probably implement this sometime soon.

amerlyq:
Yes, insterested. Also I would appreciate if 'space' password can also be supplied from stdin to skip prompt and keep secure data encrypted.

shapeshed:
This would be a great feature and let people manage credentials how they want to.

shapeshed:
Should this be closed if there are no plans to implement it?

flashcode:
It's still open and can be implemented in future; currently I lack time to implement such features.

foudfou:
In case @incertia is still interested in contributing, @flashcode could you provide him with some guidance so he can progress on #906 ?

amerlyq:
Actually, in case of nickserv we could use something like this to read password from cmd stdout
```
/set irc.server.bitlbee.command /exec -nosh -nobg -nostdin -norc -noln -pipe "/msg nickserv identify" pass irc/$server
```
But it doesn't work for `sasl_password` or SSL `password`.
However, as these fields are evaluated, we could use something like this:
```
${eval:/exec -nosh -nobg -nostdin -norc -noln pass irc/$server}
${exec:-nosh -nobg -nostdin -norc -noln:pass irc/$server}
${exec:pass irc/$serve}
```
Of course, currently it doesn't work.
But this idea is rather portable and can be used anywhere, as @flashcode was concerned in #906.

binaryplease:
Hi,

[...]

PR #906, opened 2017-02-09
irc: add the per-server password_cmd option.

incertia:
This is my initial solution to tackle #141. An alternative I considered was adding a shell type variable in the `/eval` function. I decided against this because perhaps in the future (if we deem this is a good idea), we can add support for different "passphrase plugins". For example, different plugins to interface with `gnome-keyring` or just pulling in a straight password via `pinentry`, although in hindsight I suppose those can go into eval as well.

### DIFF
--- a/src/plugins/irc/irc-command.c
+++ b/src/plugins/irc/irc-command.c
@@ -4588,6 +4588,14 @@ irc_command_display_server (struct t_irc_server *server, int with_detail)
             weechat_printf (NULL, "  password . . . . . . : %s%s",
                             IRC_COLOR_CHAT_VALUE,
                             _("(hidden)"));
+        /* password_cmd */
+        if (weechat_config_option_is_null (server->options[IRC_SERVER_OPTION_PASSWORD_CMD]))
+            weechat_printf (NULL, "  password_cmd . . . . :   %s",
+                            _("null"));
+        else
+            weechat_printf (NULL, "  password_cmd . . . . : %s%s",
+                            IRC_COLOR_CHAT_VALUE,
+                            weechat_config_string (server->options[IRC_SERVER_OPTION_PASSWORD_CMD]));
         /* client capabilities */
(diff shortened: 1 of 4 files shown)


flashcode:
Hi,
Thanks for the PR.
I think it's better to have a global solution rather than an IRC specific one, so that other plugins, scripts or even core of weechat can benefit of that (many scripts are using/handling passwords for external services).
I think something integrated with /secure feature (secured data) would be nice.

incertia:
With respect to that, here is something that could work. We move the existing values `sec.data.var` to `sec.data.var.value`, as well as adding the suboption `sec.data.var.plugin`, which takes strings to identify how `value` should be interpreted. e.g. `plain` for plaintext (the default), `shell` for a shell command, `gnome-keyring` for information related to the keyring and key to extract, etc.

This would require changes in both eval and secure, which I am happy to make. Ideally, we should not run these plugins until the values are actually used (hence eval) to avoid keeping the password in memory.

flashcode:
Hi,

Thanks for your PR.

I'm closing it because I implemented a way to read the passphrase with an external program, see: https://github.com/weechat/weechat/issues/141#issuecomment-774487254
So the secured data itself is still stored in `sec.conf` and handled by WeeChat itself.

Reading secured data itself with an external program is hard to implement without major refactoring, it must be asynchronous and not block WeeChat, but in many cases we need the password immediately (example: to connect to an IRC server).

I hope this first step with the passphrase is enough for most users.

GideonWolfe:
Is there any plans to add per-server password command evaluation to weechat? I'm using NixOS and Agenix to manage secrets on my OS, and I have fully decrypted secret files available to be read on weechat startup.

My goal is to have my configuration fully managed in text with as little manual setup as possible, and at this point having to manually set up the crypt every time just seems like unnecessary work. It would be much nicer to simply save the per-server password in something like `sec.crypt.<server>.passwordFile` or even `sec.crypt.<server>.passwordCmd`.

PR #907, opened 2017-02-09
doc: fix examples with localvar_del

haavard:
The docs provide the following example for deleting localvars on buffers:

    /* remove local variable "toto" */
    weechat_buffer_set (my_buffer, "localvar_del_toto", NULL);

This does not work, because [`gui_buffer_set`][1] exits immediately if one of the arguments are `NULL`. This PR simply updates the documentation to reflect this, providing the following example instead:

    /* remove local variable "toto" */
    weechat_buffer_set (my_buffer, "localvar_del_toto", "");

[1]: https://github.com/weechat/weechat/blob/master/src/gui/gui-buffer.c#L1795


flashcode:
Thanks for the fix.

Issue #320, opened 2015-01-24
Make irc.server_default.capabilities like weechat.plugin.autoload & make * safe capabilities that break nothing


Mikaela:
Currently WeeChat requires all capabilities enabled manually in `irc.server_default.capabilities`. WeeChat should automatically request safe capabilities where safe means that it shouldn't break anything (possibly external to WeeChat like pisg).

Combining comments there could be `*` as special capability that requests the safe ones and it could be configured like `weechat.plugin.autoload`.

> `comma separated list of plugins to load automatically at startup, "*" means all plugins found, a name beginning with "!" is a negative value to prevent a plugin from being loaded, wildcard "*" is allowed in names (examples: "*" or "*,!lua,!tcl")`

The default value would be `*`.

The safe capabilities could be:
- account-notify
- away-notify
- cap-notify
- multi-prefix
- server-time & znc.in/server-time-iso (1.7.0+ support server-time (https://github.com/znc/znc/commit/d7a6a136dbdd04af2c9f403a362f63d010fb00af), but people and distributions are SLOW to upgrade.)
- znc.in/self-message (https://github.com/weechat/weechat/commit/b860043646755773a2d6173d411f80e2c3b89182)

_This issue was originally on requesting all capabilities, but it has been decided to be possibly a very bad idea™ and I rewrote this on 2015-03-30. I also removed all my comments as they were used to derail from this issue._

stfnm:
Maybe allowing a wildcard value like "*" for this option meaning "all supported capabilities" would solve this nicely.

Mikaela:
```
2015-01-24 16:27:38+0200 < grawity> the .capabilities setting could maybe allow cap1,cap2,!cap3 syntax in case someone wants to forcefully disable a known capability
2015-01-24 16:27:51+0200 <@FlashCode> yes like plugins
2015-01-24 16:27:57+0200 <@FlashCode> something like "*" for automatic
2015-01-24 16:28:18+0200 <@FlashCode> or empty string (or "-") for nothing by default
```

Shawn-Smith:
I wouldn't mind seeing `*` for "all capabilities", but I don't think they should all be enabled by default.

Mikaela:
>  I don't think they should all be enabled by default.

and why is this?

Shawn-Smith:
Because some capabilities people don't need, like `away-notify` is only useful if you use the `away_check` setting. And the list of capabilities that aren't needed will only grow as more time passes.

[...]

PR #919, opened 2017-03-05
preliminary fix for issue #320

rhg:
I'm not sure exactly on the worth of this approach, but it's useful for now so I implemented it.

I did a bit of testing and it seems to work.


flashcode:
As #320 has been implemented (in WeeChat 3.3), this PR can be closed.

PR #929, opened 2017-03-21
doc: add weechat.look.align_multiline_words to FAQ, remove unnecessary info, fix typos/grammar

arza-zara:
* Added info about weechat.look.align_multiline_words introduced in 1.7.
* Removed info on timestamp and prefix alignment which is already covered in [#customize_prefix](https://weechat.org/files/doc/weechat_faq.en.html#customize_prefix) and isn't so URL specific.
* Fixed typos/grammar.


flashcode:
Merged, thanks.

PR #931, opened 2017-03-21
Add weechat.look.prefix_same_nick_more

### LINKED ISSUE #930: Improve prefix_same_nick with a special prefix for middle lines
I've long wanted a different prefix for the middle line between two messages sent by the same user, as I've always thought it would look good.

Screencap of my implementation:

abjugard:
Defines a prefix to be displayed on the middle lines if same nick sends
more than 2 messages in a row

This fixes #930.


flashcode:
Thanks (and sorry for the long delay until the merge!).

Issue #916, opened 2017-02-25
cmake: Inform user about missing libraries


sideeffect42:
When compiling WeeChat with CMake it may silently disable explicitly enabled options when dependencies are missing.

E.g. when `ENABLE_PERL=ON` is set, but `libperl-dev` is not installed, WeeChat will silently not compile the perl target.

Instead there should be a warning or error, so that the user can take action (install dependencies or disable options)

flashcode:
It will be closed when #956 will be merged.

PR #956, opened 2017-04-19
CMake: Warn (or error) about missing dependency

Neui:
With this change, when an dependency can't be build because it is missing something, it prints out an message (`Couldn't find NAME!` or `Couldn't find NAME! Required for THING.`).

If `STRICT` is set to `ON`, it'll send an error, kinda like other projects. It won't abort the process so the user can see further errors (but they won't be able to build). By default, it is set to `OFF` so existing builders won't break.

Also, an function `not_found_msg` is introduced for printing the message and determining whenever it should send an error. It is declared in the main `CMakeLists.txt` because it is small, but I'm unsure if it should go to an module.

This probably needs some more testing. I hope I covered all cases.


arza-zara:
Closes #916.

flashcode:
Done, with a slightly different implementation.

PR #965, opened 2017-04-25
irc: start from the beginning of nicks at automatic reconnection (closes #337)

### LINKED ISSUE #337: When connecting, always try preferred nicks in order
If the first preferred nickname for a server is in use when weechat connects, the second one will be used, but if weechat is subsequently disconnected and automatically attempts to reconnect, it will skip the first nickname and try the third if the second is in use.

arza-zara:
See #337.


flashcode:
Thanks, merged!

PR #971, opened 2017-04-27
Search WEECHAT_EXTRA_LIBDIR for plugins

lheckemann:
In addition to searching the statically configured WEECHAT_LIBDIR
(weechat's installation directory) for plugins, search the path
given in the environment variable WEECHAT_EXTRA_LIBDIR. This makes
departing from the FHS standard while keeping the plugins packaged
separately easier. This change was made specifically with the Nix
package manager in mind, but can easily be used by others.


flashcode:
Merged, with some changes due to recent commits, thanks.

Issue #899, opened 2017-01-29
Can't scroll up in bare mode if switched to bare mode at the top of the buffer


arza-zara:
1. alt+home
2. alt+l
3. pagedown
4. pageup -> doesn't work

PR #978, opened 2017-05-02
Search & scroll fixes

arza-zara:
Fixes for the following:
- filter indicator showed in day change message (in screenshot)
- searching fails when there are matches only after the starting point, or only before the starting point in free content buffers (alt+home, ctrl+r, type) (in screenshot)
- first letter in search mode scrolls up even if the first line matches
- searching previous match at the top of a free content buffer scrolls to the bottom (ctrl+r, b, alt+home, up)
- searching next match at the end of the buffer scrolls to the first match (ctrl+r, b, alt+end, down)
- /window scroll +n at the end of the buffer scrolls to the top
- /input's completion, buffer switch and search functions are disabled in search mode (a leftover for when scroll state wasn't saved?)
- scrolling up in bare mode fails when switched to bare mode at the top of the buffer (#899)
sc

### DIFF
--- a/src/gui/curses/gui-curses-chat.c
+++ b/src/gui/curses/gui-curses-chat.c
@@ -1875,6 +1875,9 @@ gui_chat_draw_free_buffer (struct t_gui_window *window, int clear_chat)
             ptr_line = gui_line_get_next_displayed (ptr_line);
         if (ptr_line)
         {
+            window->scroll->first_line_displayed =
+                (ptr_line == gui_line_get_first_displayed (window->buffer));
+
             y_start = (window->scroll->start_line) ? ptr_line->data->y : 0;
             y_end = y_start + window->win_chat_height - 1;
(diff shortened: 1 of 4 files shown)


src/gui/curses/gui-curses-chat.c
```diff
@@ -1873,6 +1875,9 @@ gui_chat_draw_free_buffer (struct t_gui_window *window, int clear_chat)
             ptr_line = gui_line_get_next_displayed (ptr_line);
         if (ptr_line)
         {
+            window->scroll->first_line_displayed =
+                (ptr_line == gui_line_get_first_displayed (window->buffer));
+
```
flashcode: I fixed this problem in the issue #899, commit 5d4220a91f8aa1467302dac2e7a4bfdb7a017009, so the commit in this issue for the bare display scrolling issue could be removed.
By the way, I think your fix was not working properly: it always sets first_line_displayed to 0 because the variable should be set earlier (before display).

Issue #992, opened 2017-05-17
Option to prevent resetting read marker and hotlist when switching buffers


devkev:
There are times when I switch to a buffer (since it has shown up in my hotlist), only to discover that for some reason, I now need to switch away again.  For example, I don't have time to deal with the messages in the buffer right now, but would like to return to them later.

Unfortunately, when I switch away from the buffer the read marker and hotlist are reset, which means that the buffer no longer shows up in my hotlist.  This makes it highly likely that I will forget to return to it later, as I originally intended.

I tried to solve this by using multiple windows, but it is a clumsy workaround - eventually I have too many windows open.

I figured out that what I really wanted was to have fully manual control over when the read marker and hotlist for a buffer are cleared.  This would allow me to "preview" or "peek" at a buffer, knowing that if it turns out that I don't have time to respond or deal with the messages, I can safely navigate away from the buffer without losing the read marker position and hotlist counts.

To this end, I need a way of inhibiting Weechat's currently non-configurable behaviour of resetting the read marker and hotlist when I switch away from a buffer.  If this is possible, then I can simply bind a key to `/input  set_unread_current_buffer` and press that when I would like to reset the buffer's read marker and hotlist.

weechatter:
There is a script to set buffer as unread. Maybe its possible to set the read_marker in any way

devkev:
`unread_buffer.py` only sets the hotlist.  I remember looking around in the source and not finding any way to set the read marker to an arbitrary location in the buffer, even via the plugin API.  But I could be wrong or misremembering.

In any case, using a command (or buffer switch trigger) to put the read marker and hotlist back to the way they were prior to switching away, seems like a poor solution to this problem (eg. it invites races).

mkoskar:
FYI. There is also #552 which is in part similar (not resetting read marker automatically).

PR #993, opened 2017-05-17
Add weechat.look.read_marker_update_on_buffer_switch

devkev:
AFAICT, it's not possible to achieve what I'm looking for in #992 just via the configuration and/or plugins.

I identified the places in the code where this behaviour is occurring (non-configurably), and confirmed that commenting out these lines was enough to provide the behaviour that I was looking for.  I then added an option and cleaned up the code, so that the existing behaviour is preserved by default, but interested users can still easily opt-in to this alternate behaviour.

I have been using this on a daily basis for the past 2 months, and haven't noticed any ill-effects.  I find that this approach of manually resetting the read marker and hotlist fits nicely with my workflow.

I believe that the code changes should be fairly straightforward and low-risk.  Please let me know if there is anything that can be done to improve this PR.

### DIFF
--- a/src/core/wee-config.c
+++ b/src/core/wee-config.c
@@ -135,6 +135,7 @@ struct t_config_option *config_look_hotlist_short_names;
 struct t_config_option *config_look_hotlist_sort;
 struct t_config_option *config_look_hotlist_suffix;
 struct t_config_option *config_look_hotlist_unique_numbers;
+struct t_config_option *config_look_hotlist_update_on_buffer_switch;
 struct t_config_option *config_look_input_cursor_scroll;
 struct t_config_option *config_look_input_share;
 struct t_config_option *config_look_input_share_overwrite;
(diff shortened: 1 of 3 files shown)


mkoskar:
@devkev Nice patch! I was looking for exactly this having very similar usecase as you've described in #992. Let me just add few comments though.

1) '/input jump_smart' will still remove target buffer from hotlist, I suppose that should be inhibited too?
2) shouldn't name of the option reflect better that it's not only about updating (or not) of read marker but also about removing (or not) of buffer from the hotlist? maybe having 2 separate options?

devkev:
Sorry for the delay.

> 1. `/input jump_smart` will still remove target buffer from hotlist, I suppose that should be inhibited too?

I've never used this, but I guess so?  Unless that behaviour means that this command always jumps to the same buffer.

> 2. shouldn't name of the option reflect better that it's not only about updating (or not) of read marker but also about removing (or not) of buffer from the hotlist? maybe having 2 separate options?

Agreed.

I like your `weechat.look.read_marker_auto_set` in #1008, although would you consider `read_marker_auto_clear` instead?

For the hotlist, how about a matching `weechat.look.hotlist_auto_clear` setting?

I'm not sure when I might get time to actually make these changes, so if you would like to do it in #1008, please go for it.

devkev:
Rebase against current master is clean.  This patch applied to 3.0 still works fine for my purposes.

flashcode:
@devkev: thanks for the update, I should merge this soon.
What about splitting the option into 2 options? For me the read marker and the removal of buffer from hotlist are two different things.

devkev:
Sure, no problem.  I understand that for a patch like this, the main thing is getting the choice of options right, since changing them later is a pain.  How about this (with suggested squashed commit msg)?

flashcode:
Thanks, merged and squashed the two commits.

Issue #552, opened 2015-10-17
add new "sticky" command to move the read marker


highflyer77:
Have a command to move the read marker to somewhere else in the chat area and make it sticky, 
so that it doesn't reset when switching buffers.  That would make it easier to catch up on reading,
while still being able to switch buffers.

Petzku:
This could be very useful, for instance when reading backlogs (say, something over 100 lines) and someone messages you, so you don't have to remember exactly where you were, or just quickly read a highlight in the same buffer, then scroll back up to where you were.

mkoskar:
I would suggest to change `/input set_unread_current_buffer` such that it would place read marker after last visible line at the bottom of the current window and scroll one line to make marker visible. Or better yet:

* `/input set_unread_current_buffer` - scroll to end of buffer (this **doesn't** happen now) and place read marker at the end
* `/input set_unread_current_buffer_here` - place read marker at current scrollback position (as described above)

Above naming and behavior is more or less consistent with `search_text` / `search_text_here` or `search_stop` / `search_stop_here`.

Furthermore configuration option e.g., `weechat.look.read_marker_auto {on|off}` could designate whether automatic placing of read marker at the end of buffer should take place when buffer is no longer displayed in any window.

dstensnes:
Yes please, i have been trying to figure out if there is a way of preventing it from automatically moving the read marker. +1 from me

ksu-0:
I think a good solution could look like this:
* after switch to buffer, if it is not scrolled and the read marker is not in view, scroll up until it is at the top of the view
* before switch from buffer, if it is scrolled and the read marker is located before the view, move the read marker to the top of the view
* on switch from buffer, if it is scrolled, do not update the read marker

flashcode:
Implemented in 4.0.0 via #1895: `/buffer set unread [+/-]N`.

PR #1004, opened 2017-05-29
core: add /input set_unread_current_buffer_here

mkoskar:
**Edited:** I've force-pushed patch with few cosmetic changes, especially in relation to #1007. 

* new "/input" action:

    set_unread_current_buffer_here: set unread marker at current position

* buffer "unread" property value now means:

    "0": set unread marker before first line
    int > 0: set unread marker after end line of current scroll position
    int < 0: set unread marker after last line of buffer

    Any other value is treated as "-1".

### DIFF
--- a/doc/en/autogen/user/weechat_commands.adoc
+++ b/doc/en/autogen/user/weechat_commands.adoc
@@ -474,6 +474,7 @@ list of actions:
   grab_mouse_area: grab mouse event code with area
   set_unread: set unread marker for all buffers
   set_unread_current_buffer: set unread marker for current buffer
+  set_unread_current_buffer_here: set unread marker at current position
   switch_active_buffer: switch to next merged buffer
   switch_active_buffer_previous: switch to previous merged buffer
   zoom_merged_buffer: zoom on merged buffer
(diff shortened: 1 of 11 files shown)


mkoskar:
This is partially fixing #552.

weechatter:
something i don't like. its too limited. i suggest adding the functionality from /window scroll

mkoskar:
@weechatter I was actually thinking about it, but then it seemed lot more work and maybe really needless duplication? I mean you can still use ``/window scroll`` separately before placing marker (but only on buffer that actually has window). Only then you don't have option to set marker before ``start_line`` of scroll (though that is much easier to implement than ``end_line``). I was basically catering for use-case where you can replace ``<Ctrl-S><Ctrl-U>`` binding and mark everything visible as read, be it after end of scroll or end of buffer. I thought it would be weird to place marker before ``start_line`` although it's something that could be add easily and then combined with ``/window scroll`` you can place it rather precisely, except only for "visible" buffers (that is having window). But I have a feeling you're suggesting it because you want to move it regardless of it being displayed in window or not. Hmm, well that's certainly bigger undertaking.

PR #1008, opened 2017-05-31
core: add config_look_read_marker_auto_set

mkoskar:
This is similar to #993. It's limited only to read marker (doesn't deal with hotlist) and suggests different configuration option name (``config_look_read_marker_auto_set``) which seems more appropriate to me.


mkoskar:
Closing wrt to #993 has been merged.

PR #1012, opened 2017-06-03
Small fixes and code cleanup for problems found by CLion inspections

sim642:
I ran CLion's static analysis on weechat source, then manually went through them and fixed some that I found to be the biggest issues. All of these are very small code fixes or just cleaning up code, including:

1. Functions declared with different arguments than defined and called.
2. Local variables created only to use `sizeof` for `malloc`.
3. Missing forward declarations used in some API function arguments.
4. Broken header guard.
5. Unused `#define`-s and global variables.
6. Declared functions that are not defined and used.
7. Callbacks that are not hooked.
8. Unused `struct`s.
9. Callbacks that are no longer used.

It is easier to review these changes commit by commit, not as an entire PR. Feel free to ask any questions or exclude some changes.

### DIFF
--- a/src/core/wee-command.c
+++ b/src/core/wee-command.c
@@ -1740,7 +1740,7 @@ COMMAND_CALLBACK(debug)
 
     if (string_strcasecmp (argv[1], "color") == 0)
     {
-        gui_color_dump (buffer);
+        gui_color_dump ();
         return WEECHAT_RC_OK;
     }
 
--- a/src/core/wee-infolist.h
+++ b/src/core/wee-infolist.h
@@ -22,6 +22,8 @@
 
 #include <time.h>
 
+struct t_weechat_plugin;
+
 /* list structures */
 
 enum t_infolist_type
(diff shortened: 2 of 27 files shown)


sim642:
Requested list of unused functions (after my PR):
```
Function 'gui_bar_get_min_width' is never used
Function 'gui_bar_get_min_height' is never used
Function 'gui_buffer_visited_remove_all' is never used
Function 'gui_window_clear_weechat' is never used
Function 'irc_server_get_number_connected' is never used
Function 'irc_server_get_number_buffer' is never used
Function 'command_secure_display_data' is never used
Function 'config_file_config_insert' is never used
Function 'config_file_section_insert_in_config' is never used
Function 'proxy_search_with_option_name' is never used
Function 'secure_search_hash_algo' is never used
Function 'secure_search_cipher' is never used
Function 'upgrade_file_read_string_utf8' is never used
```

I didn't jump to immediately deleting all of them because I didn't check why they might be there. Some stuff might be added but never used by accident, some stuff might have been used in the past but isn't anymore due to other changes in the code.

flashcode:
Thanks.
I'll check unused functions and remove them if they'll not be needed in future.

PR #1014, opened 2017-06-09
trigger: if hashmap creation failed, don't use tags

Neui:
If hashmap creation fails (eg. not enough memory), it jumps to the label
"end", where it checks the pointer `tags`, that hadn't been initialized
before.

The simple fix is to initialize it before creating the hashmap.

Found by the clang static analyzer. I know, it is unlikely to happen, but let's be on the safe side.


flashcode:
Thanks, merged, but I kept the INIT macro first in funtion, because it declares local variables, so that all local variables are declared together, and are not mixed with code.

PR #1031, opened 2017-07-03
Treat channel names as valid if they are the name of a channel to which we are currently joined, regardless of actual validity per CHANTYPES

janicez:
I was having severe problems on a network that did not send 005 when using it through ZNC.

My initial fix to those problems introduced another bug where extant queries would be treated as channels. That's in the commit history.

So here you have it. A pull request designed to fix an edge case no one ever describes.


flashcode:
Hi,
I think you pushed another feature in this PR, please rebase with just the changes mentioned in the title of this PR.

janicez:
I broke it

janicez:
Do not accept, i went too far

PR #1036, opened 2017-07-09
Updated: Treat channel names as valid if they are the name of a channel to which we are currently joined, regardless of actual validity per CHANTYPES

janicez:
This pull request should fix the old PR that ran away


flashcode:
Hi,
Sorry for the long delay, but you say as comment in code that it should never happen.
So what's the use case for this change? Do you have some examples where this can happen?

janicez:
Could you point to the line number and file where this happened? Sorry, I'm suffering from caffeine-induced dementia lately.

oakkitten:
it's right [here](https://github.com/weechat/weechat/pull/1036/files#diff-cd5b67144ebacfc8950ab8969ff47941R585)

janicez:
If a server supporting whack chantypes does not publish a 005, /join could work to a target that would be treated as a PM.

The commit you specifically mention is because I accidentally treated PMs as channels.

I don't remember how this problem actually manifested

janicez:
This could also be a problem if you get SAJOINed a place.

flashcode:
Could you please provide a way to reproduce, so I can test and ensure the fix is OK in all cases?

janicez:
I don't remember why I made this PR.

janicez:
I think I was connected to a whackjob IRC server and wanted it to work properly.

flashcode:
If you have a way to reproduce the issue fixed by this PR, please let me know.
For now I just close the PR.

PR #1051, opened 2017-08-12
Add support for 64 bits time_t

danieljakots:
Hi,

I'm in the process of updating weechat on OpenBSD. In OpenBSD land, time_t is 64 bits. Here's a PR to support 64 bits time_t while being 32 bits compatible.

Cheers,
Daniel


flashcode:
Hi,
Thanks for the PR.
I checked and noticed there are many other casts of time_t to `long` in WeeChat, so I'll change them all at once.

danieljakots:
Hmm I don't see any other when I build weechat (they should create a warning AFAIK) but fixing all of them won't hurt. Thanks!

PR #1063, opened 2017-09-02
core: Add an option for multiline input in a buffer

### LINKED ISSUE #984: multiline message support
Hi, weechat is awesome and extensible, I'm using it for IRC and a bunch of other IMs. One of these is Slack through https://github.com/wee-slack/wee-slack. Slack supports sending multiline messages but I couldn't figure it out whether I can do the same in weechat. This is specially useful for code snippets, currently I can see the multiline code snippets in a single message from weechat, sadly I can't produce one. Is there a configuration for this? Would it be acceptable to request it?

https://github.com/wee-slack/wee-slack/issues/118

trygveaa:
This allows the input callback function for a buffer to receive multiple
lines at once, instead of the message being split on newline before
being sent to the callback. It adds a new flag, input_multiline, to
control this. This flag defaults to 0 which is the current behavior.

Fixes #984

### DIFF
--- a/ChangeLog.adoc
+++ b/ChangeLog.adoc
@@ -420,6 +420,7 @@ Bug fixes::
 
 New features::
 
+  * core: add flag "input_multiline" in buffer
   * core: add flag "input_get_empty" in buffer
   * core: add signals "buffer_filters_enabled" and "buffer_filters_disabled"
   * core: support loading of plugins from path in environment variable "WEECHAT_EXTRA_LIBDIR" (issue #971, issue #979)
--- a/doc/de/autogen/plugin_api/hdata.adoc
+++ b/doc/de/autogen/plugin_api/hdata.adoc
@@ -587,6 +587,7 @@ _input_callback_pointer_   (pointer) +
 _input_callback_data_   (pointer) +
 _input_get_unknown_commands_   (integer) +
 _input_get_empty_   (integer) +
+_input_multiline_   (integer) +
 _input_buffer_   (string) +
 _input_buffer_alloc_   (integer) +
 _input_buffer_size_   (integer) +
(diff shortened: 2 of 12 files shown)


MRAAGH:
Is there still something wrong with this PR? I would really like to see this happen, multiline messages exist on a lot of protocols!

trygveaa:
I'm not aware of any issues with this PR. It is working fine for me. I would love to get feedback from the WeeChat maintaners.

flashcode:
I schedule that for version 2.9.

trygveaa:
Great, thanks! I rebased in on top of master now.

trygveaa:
Btw, while this makes it possible for a script to receive the newlines you insert in the input, they are still rendered as spaces instead of newlines in the input bar.

One way to fix this is to replace `\n` with `\r` in `gui_bar_item_input_text_cb`, e.g. like this

```
diff --git a/src/gui/gui-bar-item.c b/src/gui/gui-bar-item.c
index 505b784ef..600df06a8 100644
--- a/src/gui/gui-bar-item.c
+++ b/src/gui/gui-bar-item.c
@@ -963,6 +963,8 @@ gui_bar_item_input_text_cb (const void *pointer, void *data,
         }
     }
 
+    ptr_input = string_replace (ptr_input, "\n", "\r");
+
     return ptr_input;
 }

```

@flashcode: I'm not sure if that is an acceptable solution, or if it causes issues. What do you think?

flashcode:
It looks like a hack to do that, I'm even surprised it could work (with no glitch on display).

By the way, this should be changed a bit because it causes a memory leak, the old `ptr_input` is lost when we allocate a new one.

trygveaa:
> It looks like a hack to do that

Yeah, I guess changing `gui_bar_window_draw` is better, though that seems a bit complex.

I also realized now that it would have to take `weechat.bar.input.size` into account, 

> I'm even surprised it could work (with no glitch on display).

It is the same as what `multiline.pl` does in its `input_text_display_with_cursor` hook. I haven't seen any glitches when using that, in any terminal emulator.

> By the way, this should be changed a bit because it causes a memory leak, the old ptr_input is lost when we allocate a new one.

Ah, okay.

flashcode:
Merged, thanks.

PR #1064, opened 2017-09-05
php: ensure CMake script finds PHP 7

adsr:
This should help prevent linking against PHP 5.


flashcode:
Not tested, but there was not similar problem in configure script?

adsr:
Yes, I think the pkg-config part of the configure script could compile with <7. Updated patch.

PR #1076, opened 2017-10-05
improvement: better guidance on reporting crashes

Zarthus:
* Links directly to the user guide section of report crashes
* Mention that core dumps are not enabled by default
* Overall formatting

This assumes the URL is a permanent link. if not, perhaps it'd be cool to have some sort of short-permalink like https://weechat.org/ref/report-crashes which always redirects to the right URL.


flashcode:
Thanks, merged with minor changes.

PR #1077, opened 2017-10-05
improve: docs: mention there is an option to enable crash core files

Zarthus:
For people skimming the section, it will not be immediately clear to
them, especially since everything seems to suggest this option is
enabled by default.

I've not ensured this actually renders at it should, but this looks fine based on the asciidoctor documentation.


flashcode:
Thanks, merged.

PR #1082, opened 2017-10-12
irc: Don't recreate buffer on join if it already exists, fixes #1081

### LINKED ISSUE #1081: Setting irc.look.buffer_open_before_join=on and /join-ing an already joined channel clears the nicklist
Steps to reproduce:

- Version: `WeeChat 2.0-dev (git: v1.9-344-g1ec9453b7) [compiled on Sep 28 2017 23:09:49]`
- Open up a fresh WeeChat (without configuration)
- `/set irc.look.buffer_open_before_join on`
- `/server add freenode chat.freenode.net/6697 -ssl`
- `/connect freenode`
- `/join -server freenode #weechat`
- `/buffer #weechat`
- -> All users are listed in the nicklist
- `/join #weechat`
- -> Nicklist clears, new users are added to the nicklist as they join

Fusl:


flashcode:
Hi,
Returning NULL here is considered as an error, so we shouldn't do that.
Moreover, the function irc_channel_create_buffer is sometimes called when the buffer exists (for example after a reconnection to the server with the channel still opened, but not joined).
So it's better to not call at all this function when the channel is already joined, I'll make changes accordingly.

PR #1088, opened 2017-10-20
core: allow /buffer set without value with empty default (closes #746)

### LINKED ISSUE #746: /buffer set localvar_del_* can't be conveniently used
`/buffer set localvar_del_*` should delete the corresponding buffer localvar and the respective [buffer_set](https://weechat.org/files/doc/devel/weechat_plugin_api.en.html#_buffer_set) function can be called without passing a value argument for it. The `/buffer set` command, however, requires the value argument in every case, even when it makes no sense, e.g. for `localvar_del_*` usage.

Currently possible workaround: `/buffer set localvar_del_* ""`.

sim642:


weechatter:
i still find the solution not very intuitive and inconsistent. I suggest:
/buffer del localvar   instead of    /buffer set localvar_del

sim642:
I agree this isn't very intuitive but it matches Weechat's internal usage of such buffer properties: [`buffer_set`](https://weechat.org/files/doc/devel/weechat_plugin_api.en.html#_buffer_set). The `/buffer set` command is essentially directly for manipulating those internal properties. Those include somewhat unintuitive things like `localvar_set_xxx` and `localvar_del_xxx` which actually modify buffer localvars (buffer proeprties are not localvars).

There are currently three properties which don't need a value to be "set":
* `unread`
* `key_unbind_xxx`
* `localvar_del_xxx`

This PR only changes how `/buffer set` command can be used (without value) to more closely reflect the usage of those buffer properties without giving an error about missing command arguments, which is what #746 is about.

PR #1090, opened 2017-10-24
core: add fix for ncurses with separate tinfo

alyptik:
use `pkg-config` when linking ncurses if it is available to fix the unresolved issue in PR #468 

relevant bug reports: [#41245](https://savannah.nongnu.org/bugs/?41245) [#507170](http://bugs.gentoo.org/507170) [#56000](https://bugs.archlinux.org/task/56000)


flashcode:
I tested with this fix which is adding `-ltinfo` during linking, and this breaks the display in WeeChat: the vertical lines and UTF-8 chars are rendered with letters, and accents are displayed as garbage, for example in French: "relayM- M-CM-) sur le port 9000", instead of: "relay : écoute sur le port 9000".
I'm running WeeChat inside screen on Debian Stretch. My locale is fr_FR.UTF-8.

What could be the cause of that?

PR #1094, opened 2017-11-04
Include stdio.h for FILE

coypoop:
so this header can be included even if it's not already included


flashcode:
Thanks.
The include of stdio.h was missing in other files, I committed that as well.

PR #1111, opened 2017-12-04
core: asks user for passphrase twice (closes #1109) 

### LINKED ISSUE #1109: "/secure passphrase" only asks for passphrase once
Weechat version: 1.9.1
Operating system: openSUSE Tumbleweed 20171125
Steps to reproduce: Set a passphrase using the command "/secure passphrase mysecretpassword"
Actual result: passphrase is saved immediately
Expected result: ask user to enter the passphrase a second time

When asking for the passphrase just once and writing it immediately afterwards, a user who entered it wrongly has to manually delete it from sec.conf since there is no way to remove a wrong passphrase from within weechat (You're asked to decrypt it before performing actions on it, which you can't do since you don't have the correct password...)
Most programs that handle password input ask for it twice to avoid accidental mistakes. I think weechat should do this too, compare the two entered passwords and only if they match, use it, otherwise throw an error.

RyanBruno:
This adds a new feature of asking the user to retype the passphrase they are entering to ensure it is correct. When the user runs `/secure passphrase <passphrase>` it asks them to run it again to confirm.

### DIFF
--- a/src/core/wee-command.c
+++ b/src/core/wee-command.c
@@ -5352,14 +5352,14 @@ COMMAND_CALLBACK(secure)
             }
             if (strcmp (temp_secure_passphrase, argv_eol[2]))
             {
-                temp_secure_passphrase = NULL;
                 free (temp_secure_passphrase);
+                temp_secure_passphrase = NULL;
                 gui_chat_printf (NULL, _("Passphrase does not match"));
                 return WEECHAT_RC_OK;
             }
             secure_passphrase = strdup (argv_eol[2]);
-            temp_secure_passphrase = NULL;
             free (temp_secure_passphrase);
+            temp_secure_passphrase = NULL;
             gui_chat_printf (NULL,
                              (passphrase_was_set) ?
                              _("Passphrase changed") : _("Passphrase added"));
--- a/src/core/wee-command.c
+++ b/src/core/wee-command.c
@@ -5344,7 +5344,7 @@ COMMAND_CALLBACK(secure)
         }
         else
         {
-            if (string_strcasecmp (temp_secure_passphrase, argv_eol[2]))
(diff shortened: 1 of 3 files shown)


src/core/wee-command.c
```diff
@@ -5344,7 +5344,14 @@ COMMAND_CALLBACK(secure)
         }
         else
         {
+            if (temp_secure_passphrase == NULL || strcmp (temp_secure_passphrase, argv_eol[2]))
+            {
+                temp_secure_passphrase = strdup (argv_eol[2]);
```
sim642: If `temp_secure_passphrase` was not `NULL` then it had some value that is different from `argv_eol[2]`. On this line, the pointer is overwritten and the previous `temp_secure_passphrase` string is not `free`-d, causing a leak.

src/core/wee-command.c
```diff
@@ -5344,7 +5344,14 @@ COMMAND_CALLBACK(secure)
...
+                gui_chat_printf (NULL, "Run the passphrase command again to confirm!");
```
sim642: This string should probably be translatable, i.e. `_("Run the passphrase command again to confirm!")`.

src/core/wee-command.c
```diff
@@ -5344,7 +5344,14 @@ COMMAND_CALLBACK(secure)
...
+                return WEECHAT_RC_OK;
+            }
             secure_passphrase = strdup (argv_eol[2]);
+            free(temp_secure_passphrase);
```
sim642: This `free` is only done when the user finally enters the passphrase correctly the second time. The control flow does not reach this `free` to handle the leak in described above.

Also, WeeChat style would be to have a space between `free` and `(`.

PR #1122, opened 2017-12-26
core: add detection for Ruby 2.5

jirutka:


sim642:
Could probably also use this opportunity to update ruby versions in the autotools build, which seems even more out of date.

flashcode:
Thanks.
I updated the configure script accordingly.

PR #1127, opened 2018-01-03
core: add option weechat.look.buffer_time_same

sim642:
Just like `weechat.look.prefix_same_nick` allows not repeating the nick if it's the same for consecutive lines, this PR adds an option to not repeat the displayed time if it's the same for consecutive lines. The option behaves similarly: empty string means disabled, non-empty string is evaluated (to allow colors) to use a custom string (may be a space for just hiding).

Custom strings are left-aligned and padded to the length of the displayed time. Too long custom strings are simply truncated. While it would theoretically be possible to add line prefix like alignment customization options, it is an unnecessary complication for this change, especially since it hasn't been needed so far as the time has had constant aligned width.

*Feature requested by pdx in #weechat.*

### DIFF
--- a/src/core/wee-config.c
+++ b/src/core/wee-config.c
@@ -98,6 +98,7 @@ struct t_config_option *config_look_buffer_search_force_default;
 struct t_config_option *config_look_buffer_search_regex;
 struct t_config_option *config_look_buffer_search_where;
 struct t_config_option *config_look_buffer_time_format;
+struct t_config_option *config_look_buffer_time_same;
 struct t_config_option *config_look_color_basic_force_bold;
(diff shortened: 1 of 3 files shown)


kholland950:
+1, would greatly appreciate this feature

novasenco:
so, how do get the timestamp of the message into `buffer_time_same`? Obviously, `${date:%H:%M}` doesn't work...

flashcode:
@novasenco: this is not possible, this option just allows you to display an alternate string, which does not contain the date of the message.

Feel free to add a feature request if you need to include the date of the message in this option.

Issue #1108, opened 2017-11-26
/buffer merge and unmerge don't take buffer name as argument


sim642:
Most other `/buffer` subcommands take both buffer numbers and names as argument when referring to buffers. For `/buffer merge` and `/buffer unmerge` only numbers are allowed. For consistency and convenience they could also handle buffer names as argument.

*Feature requested by GrandApe in #weechat:*
```
< GrandApe> hmm wish if /buffer merge ##channelname could work
< GrandApe> now, i need to rely on buffer numbers. doesn't scale well.
```

PR #1159, opened 2018-03-14
core: merge buffer by name

KryDos:
This is the PR for #1108

Unfortunately it covers the issue partially. I'm not really sure how it's possible to unmerge by name since buffer number is involved. It would be great if someone can help me to understand it. 

I also guess I need to update docs somewhere, but I'm not really sure where.

### DIFF
--- a/src/core/wee-command.c
+++ b/src/core/wee-command.c
@@ -792,27 +792,23 @@ COMMAND_CALLBACK(buffer)
         return WEECHAT_RC_OK;
     }
 
-    /* merge buffer with another number in the list */
+    /* merge buffer with another buffer in the list */
     if (string_strcasecmp (argv[1], "merge") == 0)
     {
         COMMAND_MIN_ARGS(3, "merge");
         error = NULL;
-        number = strtol (argv[2], &error, 10);
-        if (error && !error[0])
+
+        ptr_buffer = gui_buffer_search_by_number_or_name (argv[2]);
+
+        if (!ptr_buffer)
         {
-            ptr_buffer = gui_buffer_search_by_number ((int) number);
-            if (ptr_buffer)
-                gui_buffer_merge (buffer, ptr_buffer);
-        }
-        else
-        {
-            /* invalid number */
             gui_chat_printf (NULL,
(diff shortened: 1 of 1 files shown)


sim642:
The first line of documentation to be changed would be [here on the bottom of the same file](https://github.com/KryDos/weechat/blob/bbbb81d614ea5ee815a93d3641b5f2ef3c052803/src/core/wee-command.c#L7094) where the `/buffer` command is hooked. There the `/help buffer` format and description are defined.

I only looked deeply in the code what the `/buffer unmerge` number argument does just now and it's not similar to `/buffer merge` at all. Namely, the argument for unmerging specifies the buffer number you want the buffer to have after it's unmerged, not which buffer to unmerge (that's always the current one). Also turns out `/help` doesn't really explain the number's meaning for this subcommand either. In this interpretation taking a name argument doesn't make sense anyway. Of course this behavior could be changed but that would break compatibility which isn't a good thing. This is probably the cause for not understanding the buffer number argument for unmerging.

src/core/wee-command.c
```diff
@@ -792,27 +792,23 @@ COMMAND_CALLBACK(buffer)
...
+        ptr_buffer = gui_buffer_search_by_number_or_name (argv[2]);
+
+        if (!ptr_buffer)
         {
-            /* invalid number */
-            gui_chat_printf (NULL,
-                             _("%sError: incorrect buffer number"),
-                             gui_chat_prefix[GUI_CHAT_PREFIX_ERROR]);
-            return WEECHAT_RC_OK;
+          gui_chat_printf (NULL,
```
sim642: Indentation of this block is inconsistent.

KryDos:
hi everyone. Thanks for your feedback and sorry about messing up with the indentation. 
I've fixed indentation and updated the doc. 

As @sim642 mentioned it's not make much sense now to make unmerge command to work with buffer name. At least it's quite easy todo without breaking backward compatibility. 

So, let's keep unmerge command untouched?

KryDos:
ping

flashcode:
Thanks, merged with minor changes.

PR #1163, opened 2018-03-18
Replace getdtablesize() with sysconf(_SC_OPEN_MAX)

fornwall:
From the getdtablesize(3) man page:

       It is not specified in POSIX.1; portable applications should
       employ sysconf(_SC_OPEN_MAX) instead of this call.

Specifically, this fixes a compilation problem on Android.


flashcode:
Thanks.

Issue #1167, opened 2018-03-24
DCC GET sometimes hangs due to ack


martinetd:
Sometimes while getting a file the transfer would just stop progressing and end up timeouting and failing.
I finally managed to strace the child process while that happens and here's what I can see:
```
sendto(38, "\372R", 2, 0, NULL, 0)      = -1 EAGAIN (Resource temporarily unavailable)
nanosleep({tv_sec=0, tv_nsec=1000000}, NULL) = 0
sendto(38, "\372R", 2, 0, NULL, 0)      = -1 EAGAIN (Resource temporarily unavailable)
nanosleep({tv_sec=0, tv_nsec=1000000}, NULL) = 0
sendto(38, "\372R", 2, 0, NULL, 0)      = -1 EAGAIN (Resource temporarily unavailable)
nanosleep({tv_sec=0, tv_nsec=1000000}, NULL) = 0
sendto(38, "\372R", 2, 0, NULL, 0)      = -1 EAGAIN (Resource temporarily unavailable)
nanosleep({tv_sec=0, tv_nsec=1000000}, NULL) = 0
sendto(38, "\372R", 2, 0, NULL, 0)      = -1 EAGAIN (Resource temporarily unavailable)
nanosleep({tv_sec=0, tv_nsec=1000000}, NULL) = 0
```

basically, sendto() of only 2 means we've sent two bytes of the ack so we're retrying continuously in the loop here:
```c
int
xfer_dcc_recv_file_send_ack (struct t_xfer *xfer)
{
    int length, num_sent, total_sent;
    uint32_t pos;
    const void *ptr_buf;

    pos = htonl (xfer->pos);
    ptr_buf = &pos;
    length = 4;
    total_sent = 0;
    num_sent = send (xfer->sock, ptr_buf, length, 0);
    if (num_sent > 0)
        total_sent += num_sent;
    while (total_sent < length)
    {
        if ((num_sent == -1) && (errno != EAGAIN) && (errno != EWOULDBLOCK))
            return 0;

        /* if we can't send ACK now, just return with partial failure code */
        if (total_sent == 0)
            return 1;

        /* at least one byte has been sent, we must send whole ACK */
        usleep (1000);
        num_sent = send (xfer->sock, ptr_buf + total_sent,
                         length - total_sent, 0);
        if (num_sent > 0)
            total_sent += num_sent;
    }

    /* ACK successfully sent */
    return 2;
}
```

[...]

PR #1171, opened 2018-03-30
workaround/fix DCC get hangs

martinetd:
PR pendant of #1167.

I'll be shutting down my network now but I've hogged up quite a bit with just `TCP_NODELAY`, it works as a workaround and doesn't seem to have any obvious drawback, so we can probably use that for everyone safely.

I still don't like that the xfer logic sometimes disable intermediate acks at runtime based on retry-able error though so I'd also like to add the option even if I haven't needed it, I can remove that second commit if you think it's fluff. Given there is no message when that auto-disabling happens (and I don't see how to display something easily from the fork child) I can't really say how often it has but it feels too frequent...


martinetd:
Thanks for merging and for the style/translation/changelog fixes!

PR #1175, opened 2018-04-05
aspell: fix aspell consulting dict every keystroke

rafasc:
aspell plugin checks the dictionaries for new suggestions on every
keystroke after a misspelled word occurs. This makes text input slow,
especially when using multiple dictionaries.

If the last misspelled word remains the same use the previously stored
suggestions instead.


flashcode:
Thanks for the PR, I made some minor changes, and renamed variable "word_for_suggestions" to "misspelled_word", which makes more sense.
I also modified the behavior to be compatible with the PR #1288 which was merged yesterday.

PR #1190, opened 2018-05-13
alias: add auto-escaping special variable

### LINKED ISSUE #1140: alias: allow arbitrary text to be passed from arguments into /exec
I have a script which prints some information and lets me add a comment. I would like to be able to call this script from WeeChat using an alias, with `$*` capturing the comment to add, without manually escaping the comment's plain text. However, AFAICT there is no way to do this without mangling it in some way.

Since `/alias`'s `$*` expands the argument list directly into the command it calls, there is no point at which I can consume the text as written. `/exec -sh` will choke on apostrophes, while `-nosh`, although it handles most characters safely, _strips_ some quotes (regardless of whether the alias's `$*` has been protected by single or double quotes or none at all). Single (unescaped) quotes work only in double quotes, and double quotes only in single, so there is no quotation level I can choose for the alias which will work for all comments.

Adding a variant of `$*` to `/alias` which automatically escapes the arguments (to either shell or `-nosh` standards) would solve my problem, and seems like the best way to do so.

(Possible alternatives include changing the `-nosh` behavior—which I found surprising—and, provided `/exec`'s `-inclose` does not have a similar problem, allowing it to be used when a process is started and not just while it is running.)

foghawk:
Resolves #1140.

The new special variable `$&` expands to all arguments supplied to alias, except that double quotes are escaped with backslashes. This means that arbitrary text can be passed into (e.g.) `/exec`, with alias rather than the user handling escaping.

(I thought about using `$@`, for 'like `$*` but does something different related to quotes', but figured that might be misleading given the difference in 'something'.)

### DIFF
--- a/src/plugins/alias/alias-command.c
+++ b/src/plugins/alias/alias-command.c
@@ -272,6 +272,7 @@ alias_command_init ()
            "       $n-: arguments from 'n' to last\n"
            "      $n-m: arguments from 'n' to 'm'\n"
            "        $*: all arguments\n"
+           "        $&: all arguments, with \" replaced by \\\"\n"
            "        $~: last argument\n"
            "      $var: where \"var\" is a local variable of buffer (see "
            "/buffer localvar)\n"
--- a/src/plugins/alias/alias.c
+++ b/src/plugins/alias/alias.c
@@ -98,10 +98,11 @@ alias_search (const char *alias_name)
  */
 
 void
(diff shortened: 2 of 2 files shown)


flashcode:
Thanks, merged.

I also refactored alias replace functions to use dynamic string, making code much simple and less error-prone with the malloc and realloc functions.

PR #623, opened 2015-12-13
irc: add support for IRCv3.2 Client Capability Negotiation (closes #586)

sim642:
In addition to implementing required changes to support capabilities with values and `CAP LS`/`CAP LIST` split across multiple lines, this patch also adds client-side tracking of all capabilities. This means that weechat/plugins/scripts/triggers can easily check if some capability is available or enabled, instead of having to somehow check itself.

This also modifies `/cap` command to uppercase the subcommand which it can autocomplete because servers may require it to be uppercase.

API was changed to allow keeping track of server capabilities across upgrading so the hashtable needs to be restored from an infolist.

There's a few places with comments like `WeeChat x.y.z` which need to be updated to the proper WeeChat version this gets merged into.

_NB! This patch is in conflict with #580. I would merge this before as it's a more core change. I could modify #580 to easily merge on top of this once this gets merged._

Issue #139, opened 2014-07-25
Add support for cap echo-message


Mikaela:
Via https://github.com/znc/znc/pull/603 .

http://ircv3.net/specs/extensions/echo-message-3.2.html

If I understand correctly, this only makes server/bouncer send messages that are send from WeeChat back to WeeChat.

There is ZNC module http://wiki.znc.in/Privmsg which does this always and I have used it for months with WeeChat without any issues, so I think that WeeChat already supports this as long as WeeChat requests it from server.

Update: self-message was deprecated by echo-message.

Mikaela:
Did anyone ever do testing for echo-message? It seems that I forgot to comment long time ago about IRCv3 deprecating self-message in favour of echo-message.

Mikaela:
Quick test: it duplicates all messages you send.

BlacklightShining:
@Mikaela It doesn't _duplicate_ them, per se; it causes the server to echo them back. This is desirable as it gives clients (and users!) confirmation that a message has been received by the network and is being relayed to other users (at the expense of some bandwidth, sure, but I really can't see that being an issue in this day and age). It can also be used to test lag to the network without the use of extra `PING`s.

Personally, I would like very much to see weechat implement this capability (along with visibly marking uncomfirmed messages, using each sent message to check lag, and deferring logging for each message until confirmation is received).

Mikaela:
How Communi Sailfish does this is:

1. You send a message.
2. Server echoes the message.
3. The sent message appears in the buffer.

So the sent message is not shown at all until step 3. This is probably the easiest way to implement it, but it could be nicer such as having balls in front of user's own messages:

* Green - message delivered
* Yellow - message pending (waiting for echo)
* Red - message sending failed (better to integrate with yellow or in case there is disconnect or something?)

Echo-message is important to have so the alls can be implemented later if at all. At least two widely encountered cases have been encountered:

* If client supports server-time, but not echo-message and the client clock is on wrong time, the timestamps become inconsistent. With echo-message supported the time message was sent would be taken from server-time. (https://github.com/znc/znc/issues/1144)
* Receiving own messages from queriy playback when using a bouncer such as ZNC.

BlacklightShining:
[...]

PR #1192, opened 2018-05-16
add basic support for echo-message capability

erisdev:
this is a naive implementation that just doesn't add outgoing messages to the buffer if the capability is enabled


sim642:
PR #623 is _currently_ supposedly "in progress" to be merged for 2.2. It changes how WeeChat internally tracks capabilities (not single variables for each one but a hashtable), so this would later need to be redone to use that. Assuming @flashcode actually tries to get it into 2.2.

I remember having discussed echo-message support in greater detail a long ago while working on the rest of the IRCv3 things. I unfortunately can't recall all the arguments but such naïve implementation isn't very good from user perspective: it looks just like WeeChat has become very slow when sending messages. How is that really better?

A proper implementation that somehow displays sent/received status to outgoing messages based on echo-message would actually be beneficial. It would also use echo-message in an intentional way. It is really difficult to implement that though (no means to show message status and change it, echoed messages may appear completely different from sent ones, etc).

sim642:
Refer to #139 for tracking and deeper discussion.

PR #1193, opened 2018-05-18
irc: let "/server list" display nicks

quite:
A small tweak to conveniently display your nick along with each (connected) server in the list.

### DIFF
--- a/src/plugins/irc/irc-command.c
+++ b/src/plugins/irc/irc-command.c
@@ -5006,7 +5006,7 @@ irc_command_display_server (struct t_irc_server *server, int with_detail)
             num_pv = irc_server_get_pv_count (server);
             weechat_printf (
                 NULL,
-                " %s %s%s %s[%s%s%s]%s%s, %d %s, %d pv",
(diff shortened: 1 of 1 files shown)


src/plugins/irc/irc-command.c
```diff
@@ -5017,6 +5017,8 @@ irc_command_display_server (struct t_irc_server *server, int with_detail)
                 IRC_COLOR_CHAT_DELIMITERS,
                 IRC_COLOR_RESET,
                 (server->temp_server) ? _(" (temporary)") : "",
+                (server->nick) ? " nick: " : "",
```
sim642: This string should probably be translatable.

quite: Sure we could do that. Does it make sense to introduce a new translatable string for this change though? (i'm honestly not sure), and should that be `_(" nick: ")`? And if it should possibly be `nick:`, then the code would be a bit more invovled/ugly.

flashcode:
Thanks, merged with some changes.

PR #1208, opened 2018-06-15
Fix v8 plugin for some Linux distributions

Chaz6:
Update configure.ac to explicitly link the v8 plugin with -lpthread to fix detection and compilation on some Linux distributions.


Chaz6:
This is for CentOS 7.5..1804 x86_64. I am not sure about CMake so I will check and report back.

Chaz6:
I can confirm it is found and compiled using CMake (so no changes are needed), the problem is only with automake.

flashcode:
Thanks, merged.

Issue #1162, opened 2018-03-17
completion item will not be cleared under some circumstances


weechatter:
how to reproduce:
/set weechat.completion.partial_completion_command_arg off

/script [shift-TAB] [TAB] type anything here...

example:
/script autoload abcde
completion item shows: autoload(1) hold(1) install(2) l(2) noautoload(1) re(2) s(2) toggleautoload(1) u(3)

expected behaviour, completion item should be cleared pressing any key

reported by qgTG on freenode.#weechat

PR #1225, opened 2018-07-21
gui: fix uncleared completion item (#1162)

weechatter:


flashcode:
Could you please explain the way to reproduce the problem fixed by this PR?

weechatter:
/set weechat.completion.partial_completion_command_arg off

/script [shift-TAB]
completion item shows -> autoload(1) hold(1) install(2) l(2) noautoload(1) re(2) s(2) toggleautoload(1) u(3)

now press [TAB], the completion item will not be cleared anymore!!!
you'll have to do /redraw or you have to switch buffer to clear completion item.

PR #1228, opened 2018-07-27
don't overwrite user installed GNU auto* files

Redfoxmoon3:
autoreconf -f overwrites user installed GNU auto* files like config.sub and config.guess


flashcode:
For my information, what was the problem with the overwriting of files? What do you put manually there?

Redfoxmoon3:
If you put in updated GNU config files (config.sub and config.guess) for a new architecture/OS/whatever, and then run autoreconf -f the updated files will be overwritten with old ones

PR #1235, opened 2018-08-20
Add hotlist sort criteria "group_priority"

lgfang:
Currently weechat.look.hotlist_sort supports sorting by notify level then time
or number. This patch assigns each buffer a priority and introduces a new sort
criteria: buffer priority. This can be useful when you are more interested in
certain buffers(channels) and would like to read messages from these buffers
regardless the notify message levels.

Here is a real-world scenario when this feature helps me. When using weechat at
work, I join many buffers, some are for casual chats, some are for serious stuff.
While most messages are of the same notify level (hotlist_priority), the
messages in the buffers for serious stuff are actually more important and are
what I'd like to read first. This patch allows me to jump to corresponding
buffers first using "Alt+a" (Switch to next buffer with activity).

Below are usage examples (note that buffer priority ranges -128 ~127):
    # set sort criteria
    /set weechat.look.hotlist_sort group_priority
    # set priority of a buffer
    /set weechat.priority.python.mycompany.slack.com.&chat -10
    # alternatively, go to the designated buffer and run the following
    /buffer priority -10
    # save the configuration
    /save

### DIFF
--- a/src/core/wee-command.c
+++ b/src/core/wee-command.c
@@ -564,7 +564,7 @@ COMMAND_CALLBACK(buffer)
              ptr_buffer = ptr_buffer->next_buffer)
         {
             gui_chat_printf (NULL,
-                             _("  %s[%s%d%s]%s %s%s.%s%s%s (notify: %s)%s%s"),
+                             _("  %s[%s%d%s]%s %s%s.%s%s%s (notify: %s, priority: %s)%s%s"),
                              GUI_COLOR(GUI_COLOR_CHAT_DELIMITERS),
                              GUI_COLOR(GUI_COLOR_CHAT),
                              ptr_buffer->number,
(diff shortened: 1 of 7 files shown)


flashcode:
Hi,

I think it's a good idea to add a notion of priority for buffers in the hotlist.

But now that I'm thinking about that, I think it would be even better to change the option format for `weechat.look.hotlist_sort` to use the same format as option `buflist.look.sort` (recently added). This format is flexible to sort buffers based on their properties (and on hotlist values as well).
That way it becomes easy to group buffers in hotlist by any value, and have more than 2 criterias to sort buffers in the hotlist.
And the option could have a way to read buffer local variables, so you can sort on any custom criteria, providing you have set the local variables (using for example script buffer_autoset.py to make them persistant). This would belike the buffer "priority" you are adding in this PR.

So for example you could do something like that (exact syntax to be defined):

    /set weechat.look.hotlist_sort -hotlist.priority,-hotlist.time

With a buffer priority:

    /set weechat.look.hotlist_sort -hotlist.priority,buffer.localvar_priority,-hotlist.time

What do you think about that?

lgfang:
Hi,

Thanks for your prompt response. The feature you are thinking about appears to be much more flexible than mine. And, if it is to be implemented, my pull request is no longer needed.

Regards,
Lungang

flashcode:
Apologies, I forgot to implement, I'll start working on this very soon.

PR #1260, opened 2018-10-05
irc: use key when /join-ing in open channel buffer

sim642:
When a channel buffer is open because it was previously joined, but now has been /part-ed or kicked, the buffer stays open. Running `/join` in it joins to that open channel buffer again. This change makes that `/join` call automatically use the known channel key if it's known to weechat, e.g. was used when first joining that channel.

Really reusing the known channel key could go even further, e.g. when joining that same channel specifically by name also. This doesn't include that because that would require more complex logic in the JOIN command sending to splice in the correct keys for the correct channels and possibly reorder them to match things up properly.

Happy Hacktoberfest!

### DIFF
--- a/src/plugins/irc/irc-channel.c
+++ b/src/plugins/irc/irc-channel.c
@@ -1277,7 +1277,8 @@ irc_channel_join_smart_filtered_unmask (struct t_irc_channel *channel,
  */
 
 void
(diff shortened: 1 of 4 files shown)


flashcode:
Thanks, merged!

PR #1262, opened 2018-10-06
irc: allow specifying SSL CA per server

### LINKED ISSUE #438: Allow to specify ca certificates per network

sim642:
Closes #438.

This PR is an improvement of PR #613 (credit to @ManiacTwister), which started in the right direction but has some still unresolved issues:
1. It accesses core (only) functions from irc plugin by direct include, not via plugin API: https://github.com/weechat/weechat/pull/613/files#r50862583
2. It modifies the trust list during certificate verification instead of setting up the correct trust list before that. It seems wrong because it's not obvious why allocating a new trust list mid-verification is allowed (if it even works) and if GnuTLS doesn't break this in the future. 
3. The solution is not well extensible: other users of `hook_connect` still have no way to manipulate the trust list for their connection only.

This PR tries to solve these problems and hopefully finally bring per server SSL CAs to weechat. Every `hook_connect` user can choose to manipulate the trust list however they see fit: clear it (or not), add CAs from file (or elsewhere).

Turns out GnuTLS >=3.3.0 is required to manipulate the trust list like this. This is a significantly higher requirement than otherwise is needed but it only affects the added `ssl_ca_file` option use and should be fine on most cases (even the old Ubuntu 16.04 has GnuTLS 3.4).

Happy Hacktoberfest!

### DIFF
--- a/src/core/wee-network.c
+++ b/src/core/wee-network.c
@@ -85,6 +85,7 @@ network_init_gcrypt ()
     gcry_control (GCRYCTL_INITIALIZATION_FINISHED, 0);
 }
 
+#ifdef HAVE_GNUTLS
 /*
  * Sets trust file with option "gnutls_ca_file".
  */
@@ -92,7 +93,6 @@ network_init_gcrypt ()
 void
 network_set_gnutls_ca_file (gnutls_certificate_credentials_t gnutls_xcred)
 {
-#ifdef HAVE_GNUTLS
     char *ca_path, *ca_path2;
 
     if (weechat_no_gnutls)
@@ -110,8 +110,8 @@ network_set_gnutls_ca_file (gnutls_certificate_credentials_t gnutls_xcred)
         }
         free (ca_path);
     }
-#endif /* HAVE_GNUTLS */
 }
+#endif /* HAVE_GNUTLS */
 
 /*
  * Initializes GnuTLS.
--- a/src/core/wee-network.c
+++ b/src/core/wee-network.c
@@ -68,9 +68,6 @@
 
 int network_init_gnutls_ok = 0;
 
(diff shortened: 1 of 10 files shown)


src/core/wee-network.c
```diff
@@ -1689,9 +1671,39 @@ network_connect_with_fork (struct t_hook *hook_connect)
...
                                 GNUTLS_CRD_CERTIFICATE,
-                                gnutls_xcred);
+                                HOOK_CONNECT(hook_connect, gnutls_xcred));
+        network_set_gnutls_ca_file (HOOK_CONNECT(hook_connect, gnutls_xcred));
+        rc = (int) HOOK_CONNECT(hook_connect, gnutls_cb)
+            (hook_connect->callback_pointer,
+             hook_connect->callback_data,
+             *HOOK_CONNECT(hook_connect, gnutls_sess),
+             NULL, 0, NULL, 0, NULL,
+             WEECHAT_HOOK_CONNECT_GNUTLS_CB_INIT_XCRED);
```
sim642: Maybe this isn't the best to call the GnuTLS callbacks function when simply initializing the trust list because strictly it isn't a callback for a GnuTLS function. `INIT_XCRED` could also be added as a case under the usual `hook_connect` callback if that's more appropriate, although to me it seems more related to using GnuTLS specifically and most connections will not need to touch that part.

PR #1267, opened 2018-10-14
lua: mkdir API functions should return ERROR on error.

emkw:
All other scripting languages return ERROR, lua being the only exception.


flashcode:
Thanks for fixing such old bug (at least 10 years old).

PR #1273, opened 2018-10-23
Fix regex in relay.network.allowed_ips doc

half-duplex:
The old one, `^(123.45.67.89|192.160.*)$`, implies that the period character has no special meaning, when it actually means "any single character". This escapes periods: `^(123\.45\.67\.89|192\.160\..*)$`


flashcode:
Thanks.

PR #1288, opened 2018-12-23
aspell: set a misspelled word in buffer

weechatter:
set the misspelled word in buffer, even when there are no suggestions


weechatter:
use for example dict "en_EN" and type "kraftfahrzeug" in input_line. the word turns red, because
its not known. with this PR a misspelled word, without suggestions, will be now stored in buffer
localvar: aspell_suggest: "kraftfahrzeug"

Issue #1059, opened 2017-08-30
Away status is not restored when connecting to a ZNC that is marked away


sim642:
If marking oneself as away on a ZNC server and reconnecting, WeeChat looks like it is no longer away (e.g. no away status in "away" bar item).

ZNC sends WeeChat the 306 numeric on connection to indicate the away status: https://github.com/znc/znc/blob/34717f1092e6cc62d6a6749965a4dd3c4773731e/src/IRCNetwork.cpp#L679-L684 like:
```
:irc.znc.in 306 sim642 :You have been marked as being away
```
This 306 is not parsed by WeeChat because it has a non-standard format (only one argument is nick, while a normal 306 reply has two). WeeChat expects a reply like:
```
:irc.znc.in 306 sim642 sim642 :my away message
```
For some (to me) unknown reason ZNC does not reply in a standard way and ZNC source code advises to:
> If they want to know their away reason they'll have to whois themselves. At least we can tell them their away status...

While ZNC should probably fix themselves to supply a correct 306 in this case, WeeChat could use the described workaround to parse the differently formatted 306, set itself away and `WHOIS` itself to retrieve its own away message as well (which is needed for "away" bar item).

chrisglass:
How could I help fixing this bug? 

It is pretty much the only daily annoyance I have with weechat (well, with the interaction between weechat and ZNC).

Ideally, ZNC should indeed send a proper 306, but in the mean time, perhaps an easier first step would be to at least set the away flag when the ZNC line is detected, and leaving fetching the reason with an additional whois to a later fix - a placeholder message would be fine as a first step, IMHO. My main annoyance is forgetting that I'm away after a reattach, I don't really care what reason is (it's more of a nice to have for me).

I would be happy to help with a fix in the following ways:
- Testing (that's a given)
- Writing the patch (I would probably require some mentorship since I've not done any C++ in a long time)
- Paying the fine maintainers to write the patch. I don't think there's a bounty program set up for weechat, but I'm happy to wire some funds over somehow.

What would be the most useful contribution I could make here?

sim642:
Your comment made me have another look at this issue and it now seems to me like I might've been slightly wrong. A 306 always has the same "You've been marked as being away" message, not the away message that got set. A 301 is the one with two nicks and actually has the away message. So I now think ZNC isn't wrong and there's no other way than the suggested self-WHOIS.

[...]

PR #1302, opened 2019-02-04
irc: restore away status and message when connecting to away ZNC (#1059) 

sim642:
This PR is a fix for #1059. When numeric 306 is received (from ZNC on connection when away), WHOIS is requested on own nick using irc redirection to get the away message from numeric 301. Properly set away message is necessary for away bar item to actually show anything.

This implementation is not perfect because every 306 is followed by a WHOIS, even when using `/away` manually. Visually there's no difference but there is the added overhead of a single WHOIS request. Maybe additional conditions should be in place to only do the WHOIS when the 306 isn't the result of manual away change.

Also I did not test with `away-notify` capability which maybe makes this WHOIS trickery unnecessary.

### DIFF
--- a/src/plugins/irc/irc-protocol.c
+++ b/src/plugins/irc/irc-protocol.c
@@ -3107,6 +3107,8 @@ IRC_PROTOCOL_CALLBACK(305)
     server->is_away = 0;
     server->away_time = 0;
 
+    /* refresh localvars and bar item */
+    irc_server_set_away (server, server->nick, 0);
     weechat_bar_item_update ("away");
 
     return WEECHAT_RC_OK;
--- a/src/plugins/irc/irc-protocol.c
+++ b/src/plugins/irc/irc-protocol.c
@@ -56,6 +56,7 @@
 #include "irc-sasl.h"
 #include "irc-server.h"
 #include "irc-notify.h"
+#include "irc-redirect.h"
 
 
 /*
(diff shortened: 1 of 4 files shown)


chrisglass:
I can confirm this works with the following:

Versions:

- ZNC 1.7.1 (OpenBSD): IPv6: yes, SSL: yes, DNS: threads, charset: no, i18n: no, build: cmake
- Weechat built from this PR on Ubuntu 18.04 using:
  - sudo apt build-dep weechat
  - mkdir build
  - cd build
  - cmake ..
  - make
  - sudo make install

Steps to verify:

- Connect to an IRC server via ZNC
- set /away test
- Disconnect weechat from ZNC
- Reconnect weechat to ZNC

The away status "test" is restored when reconnecting.

lazaros52:
lazaros nusios skaip
lazarosnusios10@gmail.com
τηλ  0030 6909150561
  ellada Thessaloniki
   neapolιs tk.56727
	(SKG) του Διεθνούς Αερολιμένα Θεσσαλονίκης
 	Θεσσαλονίκη, GR

chrisglass:
Is there anything I can do to further help with this issue?

I'm happy to put time in to get this merged if needed (or money, or whatever) simply because that would mean not having to run a weechat compiled from a branch as my daily driver...

I'm not a big fan of +1 comments, so, sorry about that, but it's been a year+ so I feel it's not too frequent to be annoying. Apologies if it is.

flashcode:
@chrisglass: I just need time to fully test this PR.
There are currently 57 PR pending, it takes time to test them before merging.

PR #1304, opened 2019-02-15
Fix czech translation kick message

ZethJack:
original: <op> byl vykopnut <user> results in confusing message caused by mistranslation: op was kicked user
proposed fix: <op> vykopnul <user> should result in more understandeable message: op has kicked user

### DIFF
--- a/po/cs.po
+++ b/po/cs.po
@@ -9124,11 +9124,11 @@ msgstr "%s%s%s%s%s%s%s%s%s%s opustil %s%s%s"
 
 #, c-format
 msgid "%s%s%s%s has kicked %s%s%s %s(%s%s%s)"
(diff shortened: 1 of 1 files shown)


flashcode:
Thanks for the translation fix.

By the way, if you have extra time to help with Czech translations, this file is not maintained any more, so contributions are welcome.

flashcode:
Current status:

    cs.po: 1408 translated messages, 589 fuzzy translations, 267 untranslated messages.

So about 62% of translations are OK in Czech (1408 out of 2264).

ZethJack:
You're welcome. As far as other translation goes, I'll see what can be done. I don't have much experience in that regard though so it might take some learning on my part first.

flashcode:
The developer's guide gives info about translating process: https://weechat.org/files/doc/devel/weechat_dev.en.html#translations

PR #1307, opened 2019-02-18
aspell: strip trailing spaces from nick_completer (closes #1306)

### LINKED ISSUE #1306: Spellchecking applied to tab completed nicks
## Bug summary

After running the command `/set weechat.completion.nick_completer ": "`, spellchecking is now applied to tab completed nicks (first word on line). This was not the case in previous WeeChat releases.

---

- WeeChat version: 2.4
- OS, distribution and version: Ubuntu 18.04

sim642:
Aspell plugin's `weechat_aspell_string_is_nick` uses `nick_completer` to avoid showing spelling errors for nicknames in tab completed situations. Since 2.4 the `nick_completer` by default ends with a trailing space, which aspell doesn't account for.

This PR makes aspell strip all trailing spaces from `nick_completer` before considering it and fixes issue #1306. Repeatedly stripping the option string on every input modification might not be the best, so maybe some optimization should be added instead.

### DIFF
--- a/src/plugins/aspell/weechat-aspell.c
+++ b/src/plugins/aspell/weechat-aspell.c
@@ -52,6 +52,9 @@ struct t_weechat_plugin *weechat_aspell_plugin = NULL;
 
 int aspell_enabled = 0;
 
+char *aspell_nick_completer = NULL;
+int aspell_len_nick_completer = 0;
+
 #ifdef USE_ENCHANT
 EnchantBroker *broker = NULL;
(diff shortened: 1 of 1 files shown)


sim642:
Added a second commit to optimize the stripping by caching the stripped `nick_completer` in aspell plugin. It's slightly ugly but should avoid the completer string copying by `weechat_string_strip` on every single input modification. Not sure if it's a worthy optimization so feel free to merge without it.

Issue #1309, opened 2019-02-20
Modifying a filter forces refiltering of all buffers, not only affected buffers


sim642:
## Current behavior
Adding/deleting/etc a single filter calls `gui_filter_all_buffers` which essentially runs a triple nested loop over all buffers, all lines per buffer and all filters per line. This is extremely inefficient (WeeChat freezes) when many buffers are open but the filter only affects a single buffer.

This is one reason `grep_filter.py` script seems to be very slow: it handles single buffer filters but causes refiltering of everything.

## Desired behavior
Filter changes (including by `grep_filter.py`) don't freeze WeeChat for so long.

## Suggested solutions
More specific function instead of `gui_filter_all_buffers` which only loops over the buffers matched by the modified filter. In the worst case it's still all, in the best case it's 1 out of N.

PR #1312, opened 2019-02-21
core: add /filter addreplace

### LINKED ISSUE #1055: /filter addreplace for adding or replaceing existing filter
When testing a filter and you have something wrong, i.e. the regex isn't right, currently you must `/filter del` that wrong filter and then `/filter add` it again with the correction. Would it be nice to have it in a single command i.e. `/filter addreplace`. Especially since we already have `/trigger addreplace` too.

sim642:
Fixes #1055.

Contains a TODO regarding another place where issue #1309, PR #1311 filter freeing optimization should be used when both are merged.

### DIFF
--- a/src/core/wee-command.c
+++ b/src/core/wee-command.c
@@ -2314,9 +2314,18 @@ COMMAND_CALLBACK(filter)
     }
 
     /* add filter */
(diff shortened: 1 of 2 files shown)


flashcode:
Merged, thanks.

PR #1319, opened 2019-03-01
irc: quote argument to NICK

phy1729:
`/nick :)` results in the user's nick being changed to `)` rather than a server error. Use : before %s in each `NICK` command to avoid this.


flashcode:
Thanks for the fix.

Issue #733, opened 2016-05-12
Support relay listening on AF_LOCAL (AF_UNIX) sockets


nomis:
Local sockets exist as files with normal permissions, so it would be possible to listen on a socket that only the user has access to. There'd be no need for multiple users to agree on listening ports or for a relay password.

SSH now has support for Unix socket forwarding, so it would be possible to have a client connect to weechat via SSH with no additional passwords.

ilmaisin:
Could this feature also be documented?

flashcode:
@ilmaisin: it is documented: https://weechat.org/files/doc/stable/weechat_user.en.html#relay_unix_socket.

PR #1333, opened 2019-04-11
relay: UNIX socket support (implement #733)

ryan-farley:
This adds support for relay servers listening on UNIX domain sockets.
They can be created using the /relay addpath command and specifying the
un protocol option. 'path' is now used instead of 'port' in messages,
which contains either a string representation of the port or the path to
a socket; port is negative when a UNIX socket is used. Path based relays
are stored in their own config section, with existing callbacks reused
as much as possible. Documentation has been updated as well.

### DIFF
--- a/doc/en/weechat_user.en.adoc
+++ b/doc/en/weechat_user.en.adoc
@@ -3197,11 +3197,11 @@ The URI must always end with "/weechat" (for _irc_ and _weechat_ protocols).
 [[relay_unixsocket]]
 ==== UNIX domain sockets
 
-Using the protocol option "un" and the "/relay addpath" command, you can
+Using the protocol option "unix" with the "/relay add"  command, you can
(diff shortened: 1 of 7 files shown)


ryan-farley:
Thank you for the review. I've implemented all the requested changes, which do make it a bit simpler (with no new commands and a more sensible protocol option) and actually translatable (the lack of which was just stupid on my part).

flashcode:
I agree with @sim642, I'll rename "un" to "unix" and I'll do some other minor changes (mostly style).
The merge is in progress.

flashcode:
I'll rename to `unix_socket`, since `unix` seems to be a defined integer constant (I have no idea where it is defined).

ryan-farley:
Thank you for even considering this (let alone working to merge it). Regarding `un`, the constant `unix` is why I'd avoided it in the code, but `unix_socket` is more readable, yes.

PR #1339, opened 2019-04-16
quick start guide: rearrange content in section 2

msiism:
Put everything help-related first, then explain how to set options.


msiism:
I would further recommend splitting this up into two separate sections, one about online help and another one about setting options.

PR #1342, opened 2019-04-25
cmake: allow custom libdir (fix #1341)

### LINKED ISSUE #1341: libdir is hard-coded in weechat.pc
## Bug summary

The `weechat.pc` file has a hard-coded `libdir`.

On my system using `weechat-2.4`.
```
prefix=/usr
exec_prefix=${prefix}
libdir=${exec_prefix}/lib
includedir=${prefix}/include

Name: weechat
Description: Weechat plugins headers
Version: 2.4
Cflags: -I${includedir}/weechat
```

I have a 64-bit system which installs weechat libs to `/usr/lib64/weechat/` so I assume it should be `libdir=${exec_prefix}/lib64/weechat`?

`libdir` is set here.

https://github.com/weechat/weechat/blob/38535686f841c107cb95ff618f7b152af98a8ff4/weechat.pc.in#L3

And here.

https://github.com/weechat/weechat/blob/38535686f841c107cb95ff618f7b152af98a8ff4/CMakeLists.txt#L264

## Steps to reproduce

1. Build weechat using cmake with: `-DLIBDIR=/usr/lib64`
2. Look at `weechat.pc` in the build directory which doesn't reflect the correct lib directory.

## Current behavior

The `weechat.pc` file has a hard-coded `libdir`.

## Expected behavior

The `weechat.pc` file should show the correct `libdir` as defined by the `LIBDIR` cmake variable.

## Suggested solutions

It probably should be fixed here by removing `${PROJECT_NAME}`.

https://github.com/weechat/weechat/blob/38535686f841c107cb95ff618f7b152af98a8ff4/CMakeLists.txt#L62

But I think this will require changing `LIBDIR` accordingly in a lot of other places.

## Additional information

N/A

---

- WeeChat version: `weechat-2.4`
- OS, distribution and version: `Slackware64-current`

ryan-farley:
Allows for custom LIBDIR to specified with cmake builds. WEECHAT_LIBDIR
is used for a custom Weechat directory, or set from LIBDIR (as LIBDIR
used to be set).


flashcode:
Hi,

I tested with your changes, and the result weechat.pc is:

    prefix=/usr/local
    exec_prefix=${prefix}
    libdir=/usr/local/lib
    includedir=${prefix}/include

Shouldn't libdir be that instead?

    libdir=${prefix}/lib

ryan-farley:
At this moment it is still derived from `CMAKE_INSTALL_PREFIX`; would it be better to set `libdir` with something like 
```
string(REPLACE "${CMAKE_INSTALL_PREFIX}" "\${prefix}" libdir "${LIBDIR}")
```
in `CMakeLists.txt`?

flashcode:
Thanks, merged.
I added the replacement with ${prefix} as well (not mandatory, but nice to have).

Issue #1343, opened 2019-04-28
gui: only 256 color pairs supported even if ncurses supports more


sim642:
The following screenshot of `/color` buffer shows ncurses reporting support for 65536 color pairs but WeeChat's color pair allocation system still only uses 256 at most.  
_Screenshot by hexa- on #weechat:_
![](https://i.imgur.com/8k7xOsm.png)

WeeChat itself imposes this limit:
https://github.com/weechat/weechat/blob/38535686f841c107cb95ff618f7b152af98a8ff4/src/gui/curses/gui-curses-color.c#L579-L584

The likely reason this limit is set in WeeChat is that when color pairs above 255 would be used in WeeChat through ncurses's `COLOR_PAIR` attribute macro, they wouldn't work because the macro only supports color pairs 0..255: https://invisible-island.net/ncurses/man/curs_color.3x.html#h3-COLOR_PAIR.

It should be possible to use more color pairs by selecting the color pair separately from attributes: https://invisible-island.net/ncurses/man/curs_attr.3x.html#h2-NOTES.

According to ncurses documentation the support for more than 256 color pairs only exists in ncurses ABI 6 and not ABI 5, which still seems to be used with ncurses 6.1 on Ubuntu etc. Nevertheless, fixing this issue should be possible and would benefit users who have ncurses ABI 6.

PR #1345, opened 2019-05-02
gui: support up to 32767 ncurses color pairs (issue #1343)

sim642:
This PR fixes issue #1343:
1. All uses of the problematic `COLOR_PAIR` macro replaced with functions that allow up to 32767 color pairs (`short` datatype size limit in both the ncurses functions and WeeChat's color pair table).
2. Transposed the color table in `/color` buffer (and command line option `-c`) to sensibly allow showing more than 255 pairs.

For testing I used my [`colorspam.py`](https://gist.github.com/sim642/dde23725f23907bc960fce308ecda4f1) script.

Needs to be verified that the `setcchar` function with its `wchar_t*` argument is usable in all curses setups. Could be that this needs to be put behind a preprocessor flag to work on curses libraries without wide char support or something like that.

### DIFF
--- a/cmake/FindNcurses.cmake
+++ b/cmake/FindNcurses.cmake
@@ -23,14 +23,22 @@ endif()
 
 find_path(NCURSES_INCLUDE_PATH
   NAMES ncurses.h curses.h
-  PATHS /usr/include/ncursesw /usr/include/ncurses /usr/include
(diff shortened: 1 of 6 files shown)


clort81:
This fixes display of ascii-art, particularly with extended 0-98 colors!

Thank you!

PR #1349, opened 2019-05-26
core: update Japanese translations

l:


fzzx00348967xer073QBYXSH:
@l Why do you have the "l" username? It does not make sense for you to have such short name when everyone wants it like is there any reason why you have the @l username?

PR #1355, opened 2019-06-07
/AMSG and /AME have no valid purpose

jesopo:


flashcode:
Hi,

Could you please argument a bit this pull request, which removing these two default aliases?
If you don't use them, you can still remove them manually.

And you can still use the `/allchan` command, so I don't understand the problem with these aliases anyway.

jesopo:
- only ever really used for content that would be deemed spam
- accidentally using it is an intense pain when you have 130+ channels

flashcode:
Then what about the `/allchan` command?

And if you think you could use them by accident, you can remove the aliases from your config, no need to remove them for everybody.

For me the problem is not the alias, but the user. A user who want to send spam can do it easily, even without these aliases.

jesopo:
I'd argue that the very low public knowledge of the `/allchan` command makes this of significantly less risk of Bad People using it. Having a nice alias for it only encourages it's use.

RyanSquared:
> A user who want to send spam can do it easily, even without these aliases.

Perhaps, but it's also possible - and has happened quite often - for people to accidentally send an amsg unintentionally. I've unaliased it so that I don't accidentally do it either. It's less about spam prevention and more about *accidental* spam prevention.

SaberUK:
Some server implementations (InspIRCd, UnreaIRCd, etc) penalise users for using amsgs. In InspIRCd's case it defaults to killing users for sending mass-messages. Having client functionality which can easily trigger this seems non-ideal.

weechatter:
I also don't see a spam problem with the aliases.

flashcode:
Merged.

PR #1357, opened 2019-06-11
brew install does not take params any longer

jrylander:
As of 2 February 2019, Homebrew no longer supports options. https://apple.stackexchange.com/questions/125129/how-do-i-give-options-to-homebrew-install


flashcode:
Merged (as a single commit), and I removed such options in the user's guide as well (and the Polish FAQ you forgot to update).
Thanks for the PR.

PR #1375, opened 2019-07-10
buflist: add infolist "buflist"

sim642:
I added this infolist to make my experimental [go_buflist.py](https://github.com/sim642/go_buflist) script possible.

The point is to expose buffers displayed in the buflist in the same sort order to other scripts. An infolist seemed the easiest because buflist internally uses arraylists, so hdata wouldn't work. Also I just added the `buffer` pointer fields to the infolist because it's sufficient to be able to get any other information about the buffer with existing API.

Maybe there's a better way but any way to get this information in a script would be useful.


flashcode:
Thanks, merged.
I added a check of buffer pointer when adding to the infolit, just in case a buffer was closed.

PR #1377, opened 2019-07-11
Python script module stub

sim642:
The Python module stub `weechat.pyi` can be used for IDEs (e.g. PyCharm) to provide completion for `weechat` module functions and constants. It also includes function parameter names. I confirmed that the stub contains everything that `dir(weechat)` lists when evaluated from a WeeChat Python script.

The stub has been generated by the included script, which uses regexes to find this information from plugin API doc (functions) and scripting guide (constants). No type information is included.

In the future, a better method for generating or manually maintaining the stub is needed if there is going to be type information (which it ideally should).

## TODOs
- [ ] Is there a better place for the stub and its generator than `tools` directory?
- [ ] Scripting guide/plugin API docs should maybe mention the stub and link to wherever it is (`docs`? GitHub repo?).


flashcode:
I'm starting working on this, and I'll do the following changes:

- add type information for all arguments and return values in plugin API reference
- add type of constants in the scripting guide (integer/string)
- update the provided generation script to use the typing information in the output.

So this MR will not be merged as-is but is a very good start to make a complete stub file, that includes typing information.

flashcode:
Done!

Maybe this stub should be installed, I'll check where it makes sense to copy it (for make install and Debian packaging).

Issue #1156, opened 2018-03-06
Don't hardcode versions of libraries in CMake/configure files


flashcode:
The CMake and configure files used to detect libraries must not contain a hardcoded list of versions to search for, but only check a minimum version if needed (and if really needed a maximum version).

For example this should be avoided:

    if(PKG_CONFIG_FOUND)
      pkg_search_module(RUBY ruby-2.5 ruby-2.4 ruby-2.3 ruby-2.2 ruby-2.1 ruby-2.0 ruby-1.9 ruby-1.8)
    endif()

CMake files to update:

* [ ] FindGuile.cmake
* [ ] FindLua.cmake
* [ ] FindPHP.cmake
* [ ] FindPython.cmake
* [ ] FindRuby.cmake
* [ ] FindTCL.cmake

Detections to update are the same in script configure.ac:

* [ ] Guile
* [ ] Lua
* [ ] PHP
* [ ] Python
* [ ] Ruby
* [ ] Tcl

flashcode:
Done for Python via #1382.

flashcode:
Done for Ruby via 6dccc3181.

PR #1382, opened 2019-07-24
build: use pkg-config to properly link python

eli-schwartz:
relevant: #1156 -- If some minimum version of python is desired, this can be easily expressed with pkg-config.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,8 @@ before_script:
     - travis_retry sudo gem install asciidoctor
     - travis_retry sudo -H pip install --ignore-installed msgcheck pylint
     - phpenv local system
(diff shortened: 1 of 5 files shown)


eli-schwartz:
Finally figured out why this is totally broken: https://github.com/travis-ci/travis-ci/issues/5301

flashcode:
Hi,

Thanks for the PR.

I'm testing the changes, but this introduces a new behavior: if the option "enable_python2" is set, only Python 2 is used, while now there's a fallback on Python 3.
But I think that it's better to have just Python 3 by default and only Python 2 if we ask it (falling back on Python 3 is probably not the expected result from the user point of view).

flashcode:
Merged, thanks.
I made some additional changes to remove obsolete checks in configure and update docs.

PR #1396, opened 2019-08-21
Fix numeric 367 when timestamp is prefixed with character :

iwalkalone69:
Timestamp in 367 numeric may be prefixed by character : and, when it is present, timestamp is not parsed correctly and appears at 1970.

This fix ignores the character if present.


flashcode:
Thanks for the fix.

PR #1400, opened 2019-09-12
Fixed segfault during excessive evaluation.

stoeckmann:
It is possible to trigger a segmentation fault while processing
an evaluation of repeating string. On a Linux 64 bit system,
enter this (or adjust arguments for 32 bit accordingly):

/eval -n ${repeat:1073741824,----}

It will overflow an integer calculation because int instead of
size_t is used. Proper check of int limitations fixes this issue.

I haven't changed this specific piece of code to size_t because it
would crash in other parts of the code tree instead. For now, int
is a limitating factor when it comes to strings (and should be
enough for sane use cases).

Signed-off-by: Tobias Stoeckmann <tobias@stoeckmann.org>

### DIFF
--- a/src/core/wee-string.c
+++ b/src/core/wee-string.c
@@ -26,6 +26,7 @@
 #include <stdlib.h>
 #include <stdio.h>
 #include <errno.h>
+#include <limits.h>
 #include <stdarg.h>
 #include <string.h>
 #include <ctype.h>
@@ -224,6 +225,10 @@ string_repeat (const char *string, int count)
         return strdup (string);
 
     length_string = strlen (string);
+
+    if (count >= INT_MAX / length_string)
+        return NULL;
+
     length_result = (length_string * count) + 1;
     result = malloc (length_result);
     if (!result)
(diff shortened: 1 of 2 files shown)


src/core/wee-string.c
```diff
@@ -224,6 +225,10 @@ string_repeat (const char *string, int count)
         return strdup (string);
 
     length_string = strlen (string);
+
+    if (INT_MAX / length_string <= count)
```
flashcode: For better readability, I would reverse the condition, like this:

```c
if (count >= INT_MAX / length_string)
```
And this error case must be tested in the unit test there: https://github.com/weechat/weechat/blob/master/tests/unit/core/test-core-string.cpp#L300
Thanks

stoeckmann: Fixed. I have also added a unit test. Thank you for pointing out exactly where to place it. I have used INT_MAX there as well, although I doubt that weechat will run on a system where sizeof(int) is not 4 bytes.

If you prefer a const integer value, please let me know.

flashcode:
Thanks, merged.

PR #1414, opened 2019-10-06
core: update German translations

ingank:


po/de.po
```diff
@@ -1839,7 +1838,8 @@ msgstr ""
 "(Format: \"calc:xxx\")\n"
 "  14. eine Option (Format: \"file.section.option\")\n"
 "  15. der Name einer lokalen Variablen eines Buffer\n"
-"  16. ein hdata Name/Variable (der Wert wird automatisch in eine "
```
flashcode: The item number 9 is now a modifier, it seems this is not present and then items after this one are shifted.

ingank: ok i've fixed point 9 (and some other items)

Issue #115, opened 2014-07-02
Passphrase support for ssl client certificates


ysf:
When using  irc.server.irc-host.ssl_xxx with an unencrypted file/key, everything works fine. But using an encrypted .pem-file, a passphrase to decrypt is needed but isn't read and used by weechat. At the moment only unencrypted .pem-files can be used.

ysf:
Is this, by any chance, scheduled or maybe even resolved yet? I do not want to sound demanding, I'm just curious if it's on a todo-list or alike.

flashcode:
Not scheduled yet.
It is not easy to "ask" something to the user in weechat (not possible now).
One solution would be to add an option in server with the passphrase, and evaluate this option, so the value could be stored encrypted in sec.conf file (the option would contain something like "${sec.data.freenode_ssl_passphrase}").
And then when connecting to the IRC server, nothing would be asked to the user.

joaopizani:
The idea of making it an option in sec.conf seems good, I vote for it :)  A good compromise of security and ease-of-use.

bendem:
would love to see passwords from irc.conf stored in sec.conf as well.

flashcode:
@bendem: you can already do that, with the format I mentioned above. The options with passwords are evaluated, so the passwords themselves can be stored in sec.conf (since version 0.4.2).
See the user's guide for more info: https://weechat.org/doc/devel/user#secured_data

bendem:
oh neat, ignore my comment then.

ysf:
I second joaos vote, sounds like a viable solution to me too.

joaopizani:
Regarding the passphrases for SSL private key of the client certificate, you could try checking the option first (freenode.ssl_passphrase), and only if the option does not exists (or is blank), then ask...

TauPan:
I'm wondering if gnutls_x509_privkey_import2 could be used in irc-sasl.c (see http://www.gnutls.org/manual/html_node/Managing-encrypted-keys.html#Managing-encrypted-keys ) at least if a password from the configuration could be passed in.

However the documentation states that "Since GnuTLS 3.4.0 this function will utilize the PIN callbacks if any." and I wonder if that means that the client would block while gnutls asks for the passphrase.

TauPan:
I have a workaround, which bears some improvement but already works for me:

` irc-local-proxy.sh`
```sh
#!/usr/bin/env bash
HOSTPORT="$1"
CERTFILE="$2"
[...]

PR #1416, opened 2019-10-09
irc: add server option ssl_password for SSL certificate private key password (issue #115)

sim642:
Closes issue #115.

In my (limited) testing, it seems fine to pass `""` as password when the certificate's private key is not encrypted.

Happy Hacktoberfest!

### DIFF
--- a/src/plugins/irc/irc-server.h
+++ b/src/plugins/irc/irc-server.h
@@ -56,6 +56,7 @@ enum t_irc_server_option
     IRC_SERVER_OPTION_IPV6,          /* use IPv6 protocol                    */
     IRC_SERVER_OPTION_SSL,           /* SSL protocol                         */
     IRC_SERVER_OPTION_SSL_CERT,      /* client ssl certificate file          */
+    IRC_SERVER_OPTION_SSL_PASSWORD,   /* client ssl certificate key password */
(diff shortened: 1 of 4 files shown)


src/plugins/irc/irc-server.h
```diff
@@ -93,6 +93,7 @@ enum t_irc_server_option
     IRC_SERVER_OPTION_NOTIFY,               /* notify list                   */
     IRC_SERVER_OPTION_SPLIT_MSG_MAX_LENGTH, /* max length of messages        */
     IRC_SERVER_OPTION_CHARSET_MESSAGE,      /* what to decode/encode in msg  */
+    IRC_SERVER_OPTION_SSL_PASSWORD,   /* client ssl certificate key password */
```
sim642: Do new IRC server options need to be added to the end of this enum?

I did it just in case but if it's not necessary then it'd make sense to move this option up where `ssl_cert` is for better grouping. In most other places I put `ssl_password` right after `ssl_cert`.

flashcode: No, you can put the option where you want, so better move it with other SSL options.
But in irc-server.c, you must put the default value at the good position.

flashcode: In irc-server.c there's the name of option anyway now, but better move it as well to be consistent.

PR #1421, opened 2019-11-10
fix pkg-config linking and support python 3.8

eli-schwartz:
This contains two overlapping changes:

- try to handle pkg-config more idiomatically for modern cmake, since the world is full of ancient cmake we need to vendor modern cmake into weechat -- please verify this works on cmake 3.0 as I don't have a system to check this on

- make the python check attempt to find python 3.8 in a post https://bugs.python.org/issue36721 world. We're embedding python, so we need to explicitly link to it.


flashcode:
Tests of changes on various distributions:

* [ ] Debian Sid (unstable)
* [ ] Debian Bullseye (testing)
* [ ] Debian Buster (stable)
* [ ] Debian Stretch (oldstable)
* [x] Debian Jessie (oldoldstable)
* [x] Ubuntu Xenial (16.04)
* [ ] Archlinux
* [ ] Debian GNU/Hurd
* [ ] FreeBSD 12.1
* [ ] Windows Cygwin (32-bit)
* [ ] Windows Cygwin (64-bit)
* [ ] Windows Subsystem for Linux (WSL)

PR #1431, opened 2019-12-10
irc: fix compilation with GnuTLS < 3.1.0 (issue #115)

sim642:
Due to this ssl_password will be partially unused with GnuTLS < 3.1.0.
In that case an encrypted SSL client cert import will simply fail.

Even though GnuTLS 3.0.21 seems to have `gnutls_x509_privkey_import2` according to its git log, it is again missing from 3.0.22, which is weird. Based on their NEWS file, it was added in 3.1.0 so I decided to be safe and rely on that.

### DIFF
--- a/src/plugins/irc/irc-config.c
+++ b/src/plugins/irc/irc-config.c
@@ -1705,7 +1705,8 @@ irc_config_server_new_option (struct t_config_file *config_file,
             new_option = weechat_config_new_option (
                 config_file, section,
                 option_name, "string",
-                N_("password for SSL certificate's private key "
+                N_("password for SSL certificate's private key; "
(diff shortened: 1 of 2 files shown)


sim642:
@tjkirch This patch adds a conditional to make WeeChat still compile with older GnuTLS versions. Please try it out.

tjkirch:
@sim642 Worked great for me, thank you!

flashcode:
@sim642: a note should be added in the help of `ssl_password` option, to mention that it is used only if WeeChat is linked with GnuTLS ≥ 3.1.0.

sim642:
Added that note to `ssl_password` help. Wasn't sure what the best way to write it would be because the option already has a `(note: ...)` about evaluation.

PR #1433, opened 2019-12-13
api: add info "weechat_headless"

mhoran:
Add info "weechat_headless" so scripts can determine if weechat is running headless. Useful for remote interfaces that utilize push notifications only when weechat is inactive.


flashcode:
Thanks!

PR #1455, opened 2020-03-03
core: add detection for Ruby 2.7

lucaskanashiro:
During Ruby 2.7 transition in Ubuntu I tried this patch and weechat builds fine against it.


flashcode:
Thanks.

Issue #1454, opened 2020-03-02
Add scalable icon


sorairolake:
## Feature description
If a desktop environment support the SVG format, I think it is better to use this. Because, vector graphics does not degrade the quality even on HiDPI displays.

So, I suggest adding an icon in the SVG format to WeeChat.

flashcode:
Hi,
Good idea by I don't have any scalable format for the icon, which I made with GIMP.
I only have larger sizes available here: https://weechat.org/files/logo/ (the max size is 437x437).
If someone feels like making a SVG logo which is the same as the current bitmap image, contribution is welcome :)

sorairolake:
I see. I will create a scalable icon that is almost the same as PNG icons (make exactly the same quality is difficult).

flashcode:
Done via #1456.

PR #1456, opened 2020-03-06
Add scalable icon

sorairolake:
See #1454


flashcode:
Hi,
Thanks for the scalable icon, nice work!
I just notice it's a bit blurred compared to the original PNG files, but I think I can merge as-is and if you can fix that later that would be fine.

flashcode:
Thanks, merged.
And I removed the top/left shadow on the logo, I think it's better like that.

PR #1461, opened 2020-04-01
cmake: fix bugged WEECHAT_SHAREDIR definition

eli-schwartz:
configure.ac defines this to /usr/share/weechat, which means scripts can load from this directory as a fallback for $HOME/.weechat/, but cmake defines this instead as /usr/share, which means system-installed fallback scripts would need to be installed to /usr/share/python/foo.py instead, and that's completely insane!

Fix this discrepancy by creating a cmake define specifically for WEECHAT_SHAREDIR based on the datarootdir, just like autotools does it.


flashcode:
Thanks.

Issue #1462, opened 2020-04-02
OSS-fuzz integration


AdamKorcz:
I see that a project is setup on OSS-fuzz for WeeChat with no fuzzers setup. Are you interested in fuzzers implemented on the OSS-infrastructure? I can provide that fuzzers and integrate them into the OSS-fuzz project.

flashcode:
Hi @AdamKorcz,

Yes I plan to do that. As a first step I added a way to create fake IRC servers in version 2.8. This is mainly for testing purposes (unit tests), but could be used for fuzzing in future.

Help is welcome on this subject if you know OSS-Fuzz indrastructure. I looked at their docs but I never used.

txtsd:
Wow I added weechat to oss-fuzz back in 2016
https://github.com/google/oss-fuzz/commit/b49ba4d18fec2a0f5cdbf679a791b926ab8669e4#diff-ac6aa40970e8cc99e75fb86ed60f9697
I figured it would've been well integrated by now.
I see some work has been done in #1468 though.

Google-Autofuzz:
Anything we can do to help?

flashcode:
Hello,

I started to implement fuzz targets, with many core functions tested, and two integer overflows found so far: 68d87f2b808ae34927ea0f3dc525b1786ea621f0.

I'm testing and I'll submit very soon a PR in the ozz-fuzz repository with the appropriate config/scripts to install these fuzz targets.

PR #1468, opened 2020-04-06
Added a first fuzzer for several string operations.

DavidKorczynski:
Hi guys,

Cross referencing https://github.com/weechat/weechat/issues/1462 
I have implemented an initial fuzzer and integrated with OSS-fuzz. I have more in the pipeline but figured it would be nice to start small.

David

### DIFF
--- a/tests/fuzzers/string_fuzzer.cpp
+++ b/tests/fuzzers/string_fuzzer.cpp
@@ -1,0 +1,48 @@
+extern "C"
+{
+#include <stdlib.h>
+#include <stdint.h>
+#include <string.h>
+#include "src/core/wee-utf8.h"
+#include "src/core/wee-string.h"
+}
+
+extern "C" int LLVMFuzzerTestOneInput(const uint8_t *data, size_t size)
+{
+        char *new_m = (char*)malloc(size+1);
+        memcpy(new_m, data,size);
+        new_m[size] = '\0';
+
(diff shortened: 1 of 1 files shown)


DavidKorczynski:
Any news on this @flashcode ?

txtsd:
I'm eager to see this fuzzer added and improved on!

DavidKorczynski:
@AdamKorcz and I would be very happy to work on integrating fuzzing into weechat - I started it here as it would be nice to get the process going first before we commit more time to it. I also opened a PR on the OSS-Fuzz side of thing (https://github.com/google/oss-fuzz/pull/3584) but I closed it just a few hours ago since I wasn't sure if you were interested here - I can naturally reopen once we progress.

flashcode:
Hi @DavidKorczynski,

Thanks for the PR.

I recently added in WeeChat a way to create fake IRC servers, for now they are used in tests to simulate messages received from the server, see: https://github.com/weechat/weechat/blob/f3b9eb979da01293e44760ba6d68cce937fdd313/tests/unit/plugins/irc/test-irc-protocol.cpp#L293-L306

And my goal is to use that in a fuzzer. Simulating incoming IRC messages will cover:

- IRC protocol functions (interesting, CVE were found recenlty here)
- many string functions, including the ones you are testing here.

So I think it's a better approach to cover more code via the fuzzer.

PR #1471, opened 2020-04-07
Avoids the use of setlocale by using a separate snprintf implementation.

### LINKED ISSUE #1469: Ram usage grows speedily due to rapid repeated syscalls.
## Bug summary

Whilst using the buflist_hotlist as defined at https://gist.github.com/pascalpoitras/8406501/7b81f9fbe9326aafaf4931378751a223f4d89c44 on FreeBSD 12.1 with a manually built weechat 2.8 (using python 3.8) and locale settings as below:

```
LANG=en_US.UTF-8
LC_CTYPE="en_US.UTF-8"
LC_COLLATE="en_US.UTF-8"
LC_TIME="en_US.UTF-8"
LC_NUMERIC="en_US.UTF-8"
LC_MONETARY="en_US.UTF-8"
LC_MESSAGES="en_US.UTF-8"
LC_ALL=
```

weechat would repeatedly open `/usr/share/locale/en_US.UTF-8/{LC_COLLATE,LC_CTYPE}` then fstat, then mmap, then close. The result of this would be that RAM utilization skyrocketed, in minutes.

## Steps to reproduce

1. Duplicate the config linked above.
2. Set locale settings as mentioned above.
3. Open weechat and connect to multiple networks.
3.  Use top or htop or so forth to watch ram utilization.

## Current behavior

Repeated rapid syscalls involving the files cause ram utilization to skyrocket.

## Expected behavior

I believe that those 2 files need only be opened once most likely. It is my assumption that doing so will mean that the ram utilization will not skyrocket and will hover within a normative range.

## Suggested solutions

Still looking through the code but being unfamiliar with the code base I can only explain at a high level for now to only open locale related files once during the program's initialization probably.

## Additional information

I was unable to get this to duplicate on linux. valgrind does not work on FreeBSD currently. Using `truss` I was able to attach and watch syscalls on the PID of a running weechat with skyrocketing memory utilization. If necessary I can provide a `truss` output.  The four repeated syscalls were absent whilst the buflist_hotlist was hidden but present when it was shown are. The log generated by `truss` was 2.4M in size.

Running weechat with the command like `LC_COLLATE="C.UTF-8" LC_CTYPE="C.UTF-8" $weechat $args` does not exhibit the behavior defined as a bug and is in fact how I'm running weechat currently as a workaround.

---

- WeeChat version: 2.8
- OS, distribution and version: FreeBSD 12.1-RELEASE-p3 

I use my own pkg repository so the packages won't necessarily align with FreeBSD's main pkg repositories. The biggest difference between my own repositories and FreeBSD's is I used python 3.8 instead of a lower version.

sig-abrt:
setlocale is used in calc_format_result which is a function that is called
in rapid succession when one uses certain formatting functionalities and it
would seem smart to avoid using and/or depending upon setlocale for reasons
of speediness as well as possible broken implementations. (FreeBSD's
setlocale implementation as in 12.1-RELEASE-p3 is broken at this time.)

This commit fixes #1469

### DIFF
--- a/src/core/stb_sprintf.h
+++ b/src/core/stb_sprintf.h
@@ -1,0 +1,1879 @@
+// stb_sprintf - v1.08 - public domain snprintf() implementation
+// originally by Jeff Roberts / RAD Game Tools, 2015/10/20
+// http://github.com/nothings/stb
+//
+// allowed types:  sc uidBboXx p AaGgEef n
+// lengths      :  hh h ll j z t I64 I32 I
+//
+// Contributors:
+//    Fabian "ryg" Giesen (reformatting)
+//
+// Contributors (bugfixes):
+//    github:d26435
+//    github:trex78
+//    github:account-login
+//    Jari Komppa (SI suffixes)
+//    Rohit Nirmal
+//    Marcin Wojdyr
+//    Leonard Ritter
+//    Stefano Zanotti
+//    Adam Allison
+//    Arvid Gerstmann
+//    Markus Kolb
+//
+// LICENSE:
+//
+//   See end of file for license information.
+
+#ifndef STB_SPRINTF_H_INCLUDE
+#define STB_SPRINTF_H_INCLUDE
+
+/*
+Single file sprintf replacement.
+
+Originally written by Jeff Roberts at RAD Game Tools - 2015/10/20.
+Hereby placed in public domain.
+
+This is a full sprintf replacement that supports everything that
+the C runtime sprintfs support, including float/double, 64-bit integers,
+hex floats, field parameters (%*.*d stuff), length reads backs, etc.
+
+Why would you need this if sprintf already exists?  Well, first off,
+it's *much* faster (see below). It's also much smaller than the CRT
+versions code-space-wise. We've also added some simple improvements
+that are super handy (commas in thousands, callbacks at buffer full,
+for example). Finally, the format strings for MSVC and GCC differ
+for 64-bit integers (among other small things), so this lets you use
+the same format strings in cross platform code.
+
+It uses the standard single file trick of being both the header file
+and the source itself. If you just include it normally, you just get
+the header file function definitions. To get the code, you include
+it from a C or C++ file and define STB_SPRINTF_IMPLEMENTATION first.
+
+It only uses va_args macros from the C runtime to do it's work. It
+does cast doubles to S64s and shifts and divides U64s, which does
+drag in CRT code on most platforms.
+
+It compiles to roughly 8K with float support, and 4K without.
+As a comparison, when using MSVC static libs, calling sprintf drags
+in 16K.
+
+API:
+====
+int stbsp_sprintf( char * buf, char const * fmt, ... )
+int stbsp_snprintf( char * buf, int count, char const * fmt, ... )
+  Convert an arg list into a buffer.  stbsp_snprintf always returns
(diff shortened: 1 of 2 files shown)


sig-abrt:
> I'll look if I can just replace the setlocale + snprintf calls by something else that does not require such function.

Cool.

I went with the header file implementation for a few reasons that in the end came down to implementing a proper float to string would have taken around as much code anyway seemingly. Implementing a naive float to string would have lead to some corner cases dealing with NAN, +e$NUM formats, Infinity, and probably some other things. A naive multiply by 1000 and convert to int would have ran into issues with the maximum value allowed for int or long or so forth. 

I spent a few hours looking for float to string or "dtoa.c" files to borrow code from and ran into another candidate that was smaller, but ended up with 1.111e08 or similar type ordeals so I gave up that route and went for alternative printf implementations that don't handle locale. 

Definitely interested to see what you come up with. I lack ideas as to how to do it other than mentioned above.

flashcode:
Yes, not an easy problem to solve, it's hard to reproduce the exact behavior of snprintf in this case.
Maybe I'll just end up by removing the calls to `setlocale` and take care of comma and dot in the result (I know in French it's a comma, but I don't know for other languages).

So I could consider any char which is not minus or a digit is the decimal separator, and replace it with a dot.

sig-abrt:
What specifically drove me away from the solution you've mentioned was that in some locales (and this was at a cursory overview, I could be incorrect) the number 1000.01 would be printed as "1.000,01" which could make parsing a lot more difficult because albeit you find the first decimal indicator you'd still have to iterate over and shift characters back at every thousandth marker. 

Trying to determine in what manner the string should be parsed might be more difficult than that considering my knowledge as to locale is pretty much limited to "does utf8, colors, line drawing characters" work in a terminal and as of today "is locale's C related functions broken in some implementations and what locale can be set to make it act less broken".

flashcode:
I'm not sure any local would add "." as thousands separator in numbers, for me only the decimal separator can change. I can be wrong, but for now I can not find any information to be sure of that.

flashcode:
Anyway, to be sure there's no problem I can take care of that: if there are multiple separators like `1.000,01`, I can consider the last separator (here a comma) is the decimal separator and just remove all other separators (the dots) before this one.

flashcode:
The bug has been fixed by commit d78ac827e94a8c8ebedfea296e7447aa85c87fb5, so I close this PR.
Thanks for the help!

de-vri-es:
@flashcode: How will that code see the difference between a single thousand separator and a decimal separator though? For example: 1.234, is that 1234 or 1234/1000 ?

flashcode:
@de-vri-es: `1.234` can be ambigous, so I made the choice to consider it's a decimal separator in this case, so it's 1234/1000.
This shouldn't happen a anyway because the snprintf with format `%.10f` displays always 10 decimals.
So if there's only one separator, it's a decimal separator.

de-vri-es:
ah ok, and values like NaN and positive/negative infinity are out of scope I suppose?

flashcode:
@de-vri-es: no changes for NaN and positive/negative infinity, not handled.
For example a division by 0 just gives 0 as result (since the calculation of expressions has no way to return an error, by design).
Purpose is to give a quick way to calculate expressions, not to be an accurate scientific calculator ;)

de-vri-es:
fair enough :)

PR #1502, opened 2020-05-10
Insert multiple pasted lines in input, instead of sending immediately

trygveaa:
This changes the paste behavior so that instead of sending the lines immediately when pasting multiple lines, the lines are inserted into the input bar. This allows you to edit the paste before sending it. Then to send all the lines press enter.

To get this behavior, these changes have been done:
- Don't interpret keys when pasting, just insert the text to the input bar. This prevents the line from being sent when pasting text with newlines.
- Remove the final newline when pasting. This is so you don't end up with a blank line at the end of the input when pasting full lines.
- Remove option `weechat.look.paste_auto_add_newline`. Since you now always can edit pasted text before sending it, this option isn't useful anymore.
- Disable paste confirmation by default. Since pasted text isn't sent immediately, the confirmation isn't really necessary anymore. If you paste by mistake you can press ctrl-_ to undo it. I kept the option in case someone still wants it though.

### DIFF
--- a/src/gui/curses/gui-curses-key.c
+++ b/src/gui/curses/gui-curses-key.c
@@ -524,37 +524,37 @@ gui_key_read_cb (const void *pointer, void *data, int fd)
         }
     }
 
-    if (gui_key_paste_pending)
+    if (!gui_key_paste_bracketed)
     {
-        if (accept_paste)
+        pos = gui_key_buffer_search (0, -1, GUI_KEY_BRACKETED_PASTE_START);
+        if (pos >= 0)
         {
-            /* user is OK for pasting text, let's paste! */
-            gui_key_paste_accept ();
-        }
-        else if (cancel_paste)
-        {
-            /* user doesn't want to paste text: clear whole buffer! */
-            gui_key_paste_cancel ();
-        }
-        else if (text_added_to_buffer)
-        {
-            /* new text received while asking for paste, update message */
-            gui_input_paste_pending_signal ();
+            gui_key_buffer_remove (pos, GUI_KEY_BRACKETED_PASTE_LENGTH);
+            gui_key_paste_bracketed_start ();
         }
     }
-    else
+
+    if (!gui_key_paste_bracketed)
     {
-        if (!gui_key_paste_bracketed)
+        if (gui_key_paste_pending)
         {
-            pos = gui_key_buffer_search (0, -1, GUI_KEY_BRACKETED_PASTE_START);
-            if (pos >= 0)
+            if (accept_paste)
             {
-                gui_key_buffer_remove (pos, GUI_KEY_BRACKETED_PASTE_LENGTH);
-                gui_key_paste_bracketed_start ();
+                /* user is OK for pasting text, let's paste! */
+                gui_key_paste_accept ();
+            }
+            else if (cancel_paste)
+            {
+                /* user doesn't want to paste text: clear whole buffer! */
+                gui_key_paste_cancel ();
+            }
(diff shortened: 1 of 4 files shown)


src/gui/curses/gui-curses-key.c
```diff
@@ -494,13 +494,24 @@ gui_key_flush (int paste)
...
-        else
+        else if (!paste)
             key_str[0] = '\0';
         length_key_str = strlen (key_str);
 
         /* set last key used in buffer if combo buffer is empty */
         if (gui_key_grab || gui_mouse_event_pending || !gui_key_combo_buffer[0])
             last_key_used = i;
+
+        if (paste && (length_key_str == 63 || i == gui_key_buffer_size - 1)) {
```
trygveaa, line 505: Hard coding 63 here seems a bit fragile. Is there a better way to do this?

src/gui/curses/gui-curses-key.c
```diff
@@ -494,13 +494,24 @@ gui_key_flush (int paste)
...
+            gui_buffer_undo_snap (gui_current_window->buffer);
```
trygveaa, line 506: I'm not sure about the undo stuff, but it seemed like this was how it's done other places.

trygveaa:
@flashcode: I think we should agree on the user experience before I continue with this.

In my opinion, the best would be if a paste cause all the lines to be inserted into the input bar, without sending any message. If the pasted string ends with a newline, it will be trimmed so that the input bar doesn't end with a blank line. Pressing enter will send all of the lines.

This would mean that both the paste confirmation and paste_auto_add_newline isn't necessary anymore. Not sure if these should be removed, disabled by default or what?

src/gui/curses/gui-curses-key.c
trygveaa: This is changed, so the question is not relevant anymore.

src/gui/curses/gui-curses-key.c
trygveaa: This is changed, so the question is not relevant anymore.

trygveaa:
> This would mean that both the paste confirmation and paste_auto_add_newline isn't necessary anymore. Not sure if these should be removed, disabled by default or what?

I ended up removing `paste_auto_add_newline` and disabling paste confirmation by default.

oakkitten:
i ran this and it seems to work! a couple of moments though

* perhaps change the wording of `weechat.look.paste_max_lines` to read something along... “ask the user to confirm the paste if pasting more than this number lines”

* when `weechat.look.paste_max_lines` is, say, 2, and i paste `1\n2\n3\n` twice (shift-ins shift-ins ctrl-y), i get this in the input field
```
1
2
3[200~1
2
3
[201~
```

* when pasting more lines, various strange behavior occurs, e.g. this is me pasting `1\n2\n...9\n`:
ejk9aroqm1 1
(note the location of the cursor!)

* if i resize the window from the example above, the size of the input field changes dramatically from taking all space to vanishing completely, e.g. 
otdgpvunir 1

trygveaa:
> * when `weechat.look.paste_max_lines` is, say, 2, and i paste `1\n2\n3\n` twice (shift-ins shift-ins ctrl-y), i get this in the input field
>   ```
>    1
>    2
>    3[200~1
>    2
>    3
>    [201~
>   ```

This has now been fixed.

So I consider all issues relevant to this PR as fixed, since the rest of the things you commented behave the same in master too. But maybe I can look into creating a separate PR to fix those issues.

flashcode:
Merged, thanks!

Issue #1498, opened 2020-05-09
Improving multiline further


trygveaa:
## Feature description

Thanks for merging the multiline PR and fixing the input display! With this, scripts can have proper multiline support without using hacks as long as the user either has the `multiline.pl` script or is using the kitty or alacritty terminal emulators and set some options. However,  I have some thoughts on improving multiline further.

There are two ways to input multiline lines. Either by inserting a newline manually, or by pasing multiple lines.

### Inserting newline manually

This can be done by binding a key to `/input insert \n`. It would be nice to have a default key for this. Personally, I use alt+enter which is nice because it is similar to shift+enter used in many other applications (but unfortunately can't be used in terminals). However, alt+enter makes the application fullscreen in xterm (and on windows), so maybe some other key is better.

I see that you only applied multiline display for buffers that has `input_multiline` set. I think it's practical to be able to edit multiple lines even if it will be sent as multiple lines to the buffer, so I think it would be better if it was done for all buffers. If so, the binding is useful for all buffers. If you don't want to do this for all buffers, the binding should probably be a noop when `input_multiline` isn't set.

### Pasting

The kitty and alacritty terminal emulators sends multiple lines separated by newline characters, while other terminals I've tested separates them by carriage return. This means that to make paste work for multiple lines with in kitty/alacritty, you just have to bind `ctrl-J` to `/input insert \n`.

However, since this doesn't work for other terminals, it would be nice if bracketed paste could be used so that carriage return doesn't send the line when a paste occurs. Again, I think this would be useful even for buffers without `input_multiline` set, so you are able to edit the pasted lines.

If this is done, the paste confirmation isn't really that useful anymore imo, since a message isn't sent until you press enter, so it could be disabled by default, but this is subjective.

Additionally, it would be nice to trim the last newline after a paste, because if you have copied multiple full lines, you end up with a blank line at the end of the input, which is usually not what you want.

### Input bar size

To view multiple lines in the input bar you have to adjust `weechat.bar.input.size`. The most useful is to set it to 0, so it remains the same size when you only have one line, but adjusts dynamically when you insert multiple. I don't really see a reason to have it set to 1, so in my opinion, the default could just be changed.

trygveaa:
I also noticed some additional issues:
- If you try to insert multiple consecutive lines, i.e. blank lines, they are not shown in the input. However, when you send the message the script does receive them, so it's just a display issue.
[...]

PR #1503, opened 2020-05-12
core: Implement commands for operating on a single input line

trygveaa:
This changes the commands delete_beginning_of_line, delete_end_of_line,
delete_line, move_beginning_of_line and move_end_of_line to operate on
the current line instead of the whole input. The commands
delete_beginning_of_input, delete_end_of_input, delete_input,
move_beginning_of_input and move_end_of_input are added with the
previous implementations that the line commands had.

Additionally, the commands move_previous_line and move_next_line are
added which moves the cursor to the previous/next line and keeps the
horizontal position in the line.

The meta-r key is changed from delete_line to delete_input to keep the
behavior, and because you probably want to delete the whole input more
often than the line. The meta-R key is added for delete_line.

The home, end, ctrl-u and ctrl-k keys are kept to the same commands,
which means that they change behaviour. This is because having them
operate on the line is consistent with other applications (vim, zsh),
and I also think it's more practical.

These new bindings are added:

    shift-home:  /input move_beginning_of_input
    shift-end:   /input move_end_of_input
    shift-up:    /input move_previous_line
    shift-down:  /input move_next_line
    meta-R:      /input delete_line
    meta-ctrl-U: /input delete_beginning_of_input
    meta-ctrl-K: /input delete_end_of_input

Relates to #1498


src/gui/curses/gui-curses-key.c
```diff
@@ -98,19 +98,24 @@ gui_key_default_bindings (int context)
...
-        BIND(/* m-r           */ "meta-r",             "/input delete_line");
+        BIND(/* m-r           */ "meta-r",             "/input delete_input");
+        BIND(/* m-R           */ "meta-R",             "/input delete_line");
         BIND(/* ^T            */ "ctrl-T",             "/input transpose_chars");
         BIND(/* ^U            */ "ctrl-U",             "/input delete_beginning_of_line");
         BIND(/* ^Y            */ "ctrl-Y",             "/input clipboard_paste");
         BIND(/* <home>        */ "meta2-1~",           "/input move_beginning_of_line");
         BIND(/* <home>        */ "meta2-H",            "/input move_beginning_of_line");
         BIND(/* <home>        */ "meta2-7~",           "/input move_beginning_of_line");
         BIND(/* <home>        */ "meta-OH",            "/input move_beginning_of_line");
```
trygveaa, line 109: I don't know what terminal emulator produces this escape sequence (`/* <home> */ "meta-OH"`), so I haven't added the equivalent shift-home sequence. The same goes for `/* <end> */ "meta-OF"`, `/* ^<up> */ "meta-OA"` and `/* ^<down> */ "meta-OB"`. The escape sequences that I have added are the ones produced by xterm and urxvt.

trygveaa:
As an aside, I noticed that `delete_line` doesn't copy the text it deletes, like e.g. `delete_beginning_of_line` does. Is that intentional?

src/gui/gui-input.c
```diff
@@ -948,6 +948,96 @@ gui_input_delete_next_word (struct t_gui_buffer *buffer)
...
+        }
+
+        gui_input_text_changed_modifier_and_signal (buffer,
+                                                    1, /* save undo */
+                                                    1); /* stop completion */
+    }
+}
+
+/*
+ * Deletes all from cursor pos to beginning of input (default key: TODO).
```
trygveaa, line 1036: I'm not sure which keys would be suitable to use by default for `delete_beginning_of_input` and `delete_end_of_input`. Any suggestions?

trygveaa: I went with `meta` plus the same keys as for line (since ctrl and shift can't be used together). I.e. `meta-ctrl-U` for `/input delete_beginning_of_input` and `meta-ctrl-K` for `/input delete_end_of_input`.

src/gui/curses/gui-curses-key.c
trygveaa: Not relevant anymore with the new key bindings.

flashcode:
Merged!

PR #1504, opened 2020-05-13
core: Add an option to start multiline input text on a new line

trygveaa:
This does the same as the lead_linebreak option in multiline.pl. That
is, when the input contains more than one line, the first line will be
displayed beneath the previous items in the bar. This is practical
because all the lines in the input will be aligned.

Related to #1498


flashcode:
Merged!

Issue #879, opened 2017-01-05
tls: allow trusting keys instead of certificates


xnrand:
Certificate fingerprints in `irc.server.*.ssl_fingerprint` are the only option that weechat offers for trusting TLS servers besides the PKI as of writing this feature request.

Being able to trust certain _public keys_ instead of _certificates_ would be useful for IRC networks that renew expired certificates while reusing the same key pair(s).

IRC networks that follow this practice include AlphaChat (provided I understood [this page](https://www.alphachat.net/net-upgrades-certfp.xhtml) correctly, someone please correct me if I'm wrong) and freenode (which is planning to renew certificates with the same key pairs in the future).

This feature request is in no small part inspired by HTTP Public Key Pinning and the new certificate and pubkey pinning features in irssi 1.0.0 (irssi/irssi PR 557 for reference).

Since a good part of the work required to add this feature is probably specific to the TLS library, it might make sense to resolve this as the same time as #863.

maxteufel:
maxteufel/weechat@bfb3945897b114df895b8ff963a32804b400753d might be a start for anyone who wants to look into it. It works with some limitations specified in the commit message but it was only a small experiment so I'm not sure if I will work more on this.

PR #1507, opened 2020-05-13
plugins: irc: support server certificate public key fingerprints

aaronmdjones:
This allows users to pin known public key fingerprints instead of (or in addition to) certificate fingerprints, which means that the fingerprints will not change when the server's certificate is renewed (unless that is accompanied by a change of server private key). With the advent of short-lifetime certificates from the likes of Let's Encrypt, this is a very useful feature.

This is not any more insecure than a certificate fingerprint; server certificates only exist to securely bind a server name to a cryptographic identity (its public key), so that you have some degree of confidence that you are talking to the correct server, and not an impostor (MITM). If you're establishing its cryptographic identity up-front (by pinning a known-good fingerprint of it with a cryptographically-secure digest algorithm), then the certificate is, by definition, not needed. In fact, this is slightly more secure than a certificate fingerprint, because certificates are a lot more malleable; the ability to pad and inject random data to aid in e.g. a second preimage attack is much more easily achieved with an X.509 certificate than it is with a public key (because you can generate a valid private key, compute its corresponding public key, and then mess with other fields in the certificate to affect the digest instead). Also, any attack against `Hash(publickey)` implies an attack against `Hash(certificate)`, which would break the CA PKI model anyway.

Specifically, this fingerprints the entire `SubjectPublicKeyInfo` structure from the certificate, which includes not just the key itself, but also other meta-information about it (like its algorithm identifier and any associated parameters). This is the same public key fingerprinting approach taken by e.g. the HSTS and DANE TLSA standards, and also some IRC servers (such as Charybdis IRCd).

While working on this commit, I noticed a few oddities and/or deficiencies in the existing fingerprinting code, which this commit also corrects and/or improves:

- Fingerprints were scanned 1 byte (2 hexadecimal characters) at a time and then compared 1 byte at a time, instead of using efficient C standard library string comparison functions.

- Several for() loops used an unrelated variable to iterate over arrays of fingerprint variables (for setting their initial values or freeing them after).

- When displaying that the server's fingerprint has matched the client's configuration, it did not specify which fingerprint was received from the server for matching. This is relevant in the case of a comma-separated list of fingerprints.

- The code checked if a variable was non-NULL before calling free(3) on it, but free(3) is well-defined to do nothing if passed NULL.

cf. weechat/weechat#879


aaronmdjones:
(If anyone needs to test this, you can obtain a `SubjectPublicKeyInfo` fingerprint from an X.509 certificate like so):

`$ openssl x509 -pubkey -noout < foo.crt | openssl pkey -pubin -outform DER | sha256sum`

usrbinsam:
👍

src/plugins/irc/irc-server.c
```diff
@@ -4268,38 +4270,6 @@ irc_server_fingerprint_str_sizes ()
...
-        if (sscanf (&fingerprint[i * 2], "%02x", &value) != 1)
-            return -1;
-        if (value != fingerprint_server[i])
-            return -1;
-    }
-
-    /* fingerprints are the same */
-    return 0;
-}
-
```
sim642, line 4302: By removing this function you have done two things:

1. The `sscanf` with `%02x` allows fingerprint option values to use both lower- and uppercase. By using `strcmp` that is no longer the case and case starts to matter.

2. GnuTLS provides all fingerprints/hashes as raw arrays. This function intentionally compared those with a textual representation. By using `strcmp` you were forced to use `sprintf` below to turn all raw arrays into strings just for comparison. That's a slightly odd way of comparing raw byte arrays.

src/plugins/irc/irc-server.c
```diff
@@ -4313,75 +4283,188 @@ irc_server_check_certificate_fingerprint (struct t_irc_server *server,
...
+                "fingerprint");
+        }
+        else if (gnutls_pubkey_export (pk, GNUTLS_X509_FMT_DER, pkbuf,
+                     &pkbuf_len) != GNUTLS_E_SUCCESS)
+        {
+            weechat_printf (
+                server->buffer,
+                _("%sgnutls: failed to export certificate public key"),
+                weechat_prefix ("error"));
+        }
```
sim642, line 4363: This whole public key extraction might deserve to be a separate function, if one doesn't already exist in GnuTLS.

src/plugins/irc/irc-server.c
```diff
@@ -4313,75 +4283,188 @@ irc_server_check_certificate_fingerprint (struct t_irc_server *server,
...
+        else for (algo = 0; algo < IRC_FINGERPRINT_NUM_ALGOS; algo++)
```
sim642, line 4364: I think we don't use this `else for` style anywhere.

src/plugins/irc/irc-server.c
```diff
@@ -4313,75 +4283,188 @@ irc_server_check_certificate_fingerprint (struct t_irc_server *server,
...
                         server->buffer,
-                        _("%sgnutls: failed to calculate certificate "
+                        _("%sgnutls: failed to calculate public key "
                           "fingerprint (%s)"),
                         weechat_prefix ("error"),
                         irc_fingerprint_digest_algos_name[algo]);
-                    free (fingerprint_server[algo]);
-                    fingerprint_server[algo] = NULL;
                 }
+                else for (j = 0; j < size_bytes; j++)
```
sim642, line 4397: `else for` again.

src/plugins/irc/irc-server.c
```diff
@@ -4313,75 +4283,188 @@ irc_server_check_certificate_fingerprint (struct t_irc_server *server,
...
                     server->buffer,
-                    _("%s%s: not enough memory (%s)"),
-                    weechat_prefix ("error"), IRC_PLUGIN_NAME,
-                    "fingerprint");
+                    _("%sgnutls: failed to calculate certificate fingerprint "
+                      "(%s)"),
+                    weechat_prefix ("error"),
+                    irc_fingerprint_digest_algos_name[algo]);
             }
+            else for (j = 0; j < size_bytes; j++)
```
sim642, line 4441: `else for` again.

src/plugins/irc/irc-server.c
aaronmdjones: I can use `strcasecmp` to make it case-insensitive again if you like. I stand by making it easier for libc to compare them; plus, it's fewer lines, and fewer ways to break it in future.

src/plugins/irc/irc-server.c
aaronmdjones: It does exist in GnuTLS, in that `gnutls_pubkey_export` is all you get. This scaffolding around it is necessary to prepare for that export.

There is no `gnutls_x509_crt_get_*` function that will hand you the `SubjectPublicKeyInfo` structure; the closest is `gnutls_x509_crt_get_pk_rsa_raw` which hands you the modulus and exponent separately (and would hardcode us to RSA).

I can move it to a separate function on request, but they would still be heavily tied together...

src/plugins/irc/irc-server.c
aaronmdjones: People use `else if` all the time, but that isn't actually a syntactical construct.

This:

```C
if (foo)
    // ...
else if (bar)
    // ...
else if (baz)
    // ...
else
    // ...
```

Is just a lesser-indentation way of writing:

```C
if (foo)
    // ...
else
    if (bar)
        // ...
    else
        if (baz)
            // ...
        else
            // ...
```

They are the same. I don't see any reason not to do the same for `else for`; it just saves a layer of unnecessary indentation.

src/plugins/irc/irc-server.c
aaronmdjones: (see above)

src/plugins/irc/irc-server.c
aaronmdjones: (see above)

src/plugins/irc/irc-server.c
usrbinsam: Not seeing that in the style guide which the rest of the code appears to follow.

https://weechat.org/files/doc/stable/weechat_dev.en.html#coding_c_style

sim642: Sure-sure, but if you search through the codebase then there are at least 29 instances of a for loop directly in the else branch and that uses a properly curly-brace-separated block instead of `else for`, which we never use. It's completely unnecessary to introduce a new obscure and mixed style to our codebase.

src/plugins/irc/irc-server.c
sim642: It absolutely needs to be backwards compatible. We don't want to break existing configuration for some users for no reason.

src/plugins/irc/irc-server.c
aaronmdjones: Then I suggest the coding style be updated to either allow or forbid it, and I will refactor appropriately.

src/plugins/irc/irc-server.c
aaronmdjones: Resolved in cb355e4eb76ef4bc1da5a821e353b4e7c10251c5

src/plugins/irc/irc-server.c
aaronmdjones: Resolved in 73f890030199834be2ad7be2eaaa25c87132bab2

Issue #152, opened 2014-08-01
/layout able to store '/buffer hide' state


wlallemand:
It might be convenient if the /layout command was aware of the hide/unhide status of buffers and permits to save/restore this state.

jhanzo:
Any update on this feature request ?

weechatter:
no

sroracle:
I've taken a shot at implementing this in #1511.

wlallemand:
Thanks @sroracle ! I'm going to test this.

PR #1511, opened 2020-05-18
gui: store hidden status of buffers in layout (implements #152)

maxcrees:
This allows buffers to be persistently hidden between sessions of
weechat if the layout is saved to the configuration.

### DIFF
--- a/doc/en/includes/autogen_api_hdata.en.adoc
+++ b/doc/en/includes/autogen_api_hdata.en.adoc
@@ -876,6 +876,7 @@ _next_layout_   (pointer, hdata: "layout") +
 | _plugin_name_   (string) +
 _buffer_name_   (string) +
 _number_   (integer) +
+_hidden_   (integer) +
 _prev_layout_   (pointer, hdata: "layout_buffer") +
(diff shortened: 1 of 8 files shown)


sroracle:
As an additional note, this is especially useful in conjunction with a trigger like the following so one can keep buffers hidden indefinitely until they gain new activity:

```
unhide_buffer: line
               =? "${buffer.hidden} && ${displayed} && ${notify_level} >= 1"
               /1 "/buffer unhide ${buffer.number}"
```

sim642:
Just to have this documented: PR #1412 and this PR would cause a conflict if both were merged. Both add a new field in the config file lines of layouts.

sroracle:
Maybe we can combine them into one PR such that active/zoom/hidden state is stored in one variable as a set of flags.

sim642:
It's probably not worth the effort since in WeeChat's structs they're separate fields anyway, so correctly combining and splitting them is just extra effort. And also error-prone if one of those is changed in the future.

sroracle:
It would be more extensible if more things need to be stored in the layout configuration in the future though. I'm not sure what you mean by it being error-prone, though I have seen some people muck up enums on accident before...

Issue #1098, opened 2017-11-06
guile 2.2 not supported 


OfficialOxide:
OS: Arch Linux x86_64
Kernel: 4.13.9-1-ARCH
Steps to reproduce:

- Update to guile 2.2, removing guile 2.0
- Start weechat.

Expected result:
Guile plugin loads on startup.

Actual result:
``` 
WeeChat 1.9.1 [compiled on Sep 23 2017 19:44:10]
│ - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
│ Error: unable to load plugin "/usr/lib/weechat/plugins/guile.so": libguile-2.0.so.22: cannot open shared object file: No such file or directory
│ If you're trying to load a script and not a C plugin, try command to load scripts (/perl, /python, ...)
```

flashcode:
Hi,
Yes, this is a known problem (see also #988).
Now that guile 2.2 is available in Debian Sid, I'll try to fix compilation with this version.

IWriteThings:
Same issue on -

Linux i3-nuc 4.9.71-1-MANJARO #1 SMP PREEMPT Thu Dec 21 08:00:01 UTC 2017 x86_64 GNU/Linux

Looks like Manjaro has updated to guile 2.2 as well and causes issues with weechat.

jonsger:
@flashcode Any update/progress on guile 2.2?

jonsger:
That would be nice. I'm not a guile expert, but at least I could try to help :)

flashcode:
@jonsger: I pushed by branch "guile-2.2". It does not even compile with guile 2.0 and I didn't test again with version 2.2.

jonsger:
@flashcode Thanks. I took the branch and tried to build it with guile-2.2. I only need two simple adjustments to let it build.
It can be found here: https://github.com/jonsger/weechat/commit/ae815290a8bec2ce64361370916d8947e8511ee0

I did not test it yet. I'll report here on results :)

flashcode:
Hi @jonsger,

I tried your branch, it compiles fine with guile 2.2, but there's a regression on command `/guide eval`, this command should display "test" in the current buffer, but it displays nothing:

    /guile eval (display "test")

If ever you have a fix for that, I'll test and also test some guile scripts. Then if everything is OK, I'll test with guile 2.0 and merge to master branch if everything is OK.

Thank you for your help.

flashcode:
[...]

PR #1514, opened 2020-05-22
guile: fix output to buffers and re-enable /guile eval

jorams:
The commit messages should speak for themselves. This functionality broke when support for Guile 2.2 was added in #1098.

The documentation for Guile I/O Extensions is [here](https://www.gnu.org/software/guile/docs/docs-2.2/guile-ref/I_002fO-Extensions.html).


flashcode:
Thanks!

Issue #1446, opened 2020-01-30
publish official docker container


ssbarnea:
It would be extremely useful to publish a container and document its use as this became the standard multi-platform way to distribute apps. The biggest benefit is that users would not have bother to compile it locally and it would be very easy and fast to start or update.

flashcode:
What do you mean by "container"?

weechatter:
https://github.com/shymega/weechat-docker

ssbarnea:
>  This repository has been archived by the owner. It is now read-only. 

I think that the banner on the repo explains my concerns. The idea is to have an official docker image that is updated by CI on each release/prerelease. Some do push a new image on each commit to master.

Without official part, you depend on a single person, which may and will get bored of maintaing it at some point. Also there is the security risk.

flashcode:
Hi,

I'm closing this issue as I just added a new repository to build WeeChat container images: https://github.com/weechat/weechat-container.

Issue #360, opened 2015-03-15
Javascript plugin does not compile with v8 version 3.30.33.16


flashcode:
The javascript fails to compile with v8 version 3.30.33.16 and weechat v1.1-151-g9f15851, on Arch Linux:

```
[ 70%] Building CXX object src/plugins/javascript/CMakeFiles/javascript.dir/weechat-js.o
/home/flashcode/src/weechat/dev/src/plugins/javascript/weechat-js.cpp: In function ‘void weechat_js_hashtable_map_cb(void*, t_hashtable*, const char*, const char*)’:
/home/flashcode/src/weechat/dev/src/plugins/javascript/weechat-js.cpp:94:17: error: ‘New’ is not a member of ‘v8::String’
     (*obj)->Set(String::New(key), String::New(value));
                 ^
/home/flashcode/src/weechat/dev/src/plugins/javascript/weechat-js.cpp:94:35: error: ‘New’ is not a member of ‘v8::String’
     (*obj)->Set(String::New(key), String::New(value));
                                   ^
/home/flashcode/src/weechat/dev/src/plugins/javascript/weechat-js.cpp: In function ‘v8::Handle<v8::Object> weechat_js_hashtable_to_object(t_hashtable*)’:
/home/flashcode/src/weechat/dev/src/plugins/javascript/weechat-js.cpp:104:38: error: no matching function for call to ‘v8::Object::New()’
     Handle<Object> obj = Object::New();
                                      ^
/home/flashcode/src/weechat/dev/src/plugins/javascript/weechat-js.cpp:104:38: note: candidate is:
In file included from /home/flashcode/src/weechat/dev/src/plugins/javascript/weechat-js.h:24:0,
                 from /home/flashcode/src/weechat/dev/src/plugins/javascript/weechat-js.cpp:33:
/usr/include/v8.h:2609:24: note: static v8::Local<v8::Object> v8::Object::New(v8::Isolate*)
   static Local<Object> New(Isolate* isolate);
                        ^
/usr/include/v8.h:2609:24: note:   candidate expects 1 argument, 0 provided
/home/flashcode/src/weechat/dev/src/plugins/javascript/weechat-js.cpp: In function ‘void* weechat_js_exec(t_plugin_script*, int, const char*, const char*, void**)’:
/home/flashcode/src/weechat/dev/src/plugins/javascript/weechat-js.cpp:196:32: error: ‘New’ is not a member of ‘v8::String’
[...]

PR #1524, opened 2020-06-13
docker: add dockerfile

yancyribbens:
This PR adds a Dockerfile to the root directory of the project allowing source builds to be reproducible and isolated in a container: https://github.com/weechat/weechat/issues/1446.  I used [Alpine linux](https://alpinelinux.org/about/) for the base image because it is small and security focused: (binaries are compiled using PIE to prevent stack smashing).  Some script dependencies are difficult to add to Alpine such as V8, so I've left some script dependencies disabled for now in the Dockerfile.  We could use `debian:slim` as the base image instead which would be easier to add script dependencies I think.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,31 @@
+FROM debian:stable-slim
+
+RUN apt-get update && apt-get install apt-utils \
+    libgcrypt20-dev \
+    cmake \
+    gnutls-dev \
+    zlib1g-dev \
+    libcurl4-gnutls-dev \
+    libncurses-dev \
+    libaspell-dev \
+    pkg-config \
+    php7.3-dev \
+    python3-dev \
+    libperl-dev \
+    liblua5.2-dev \
+    ruby-dev \
+    guile-2.2-dev \
+    tcl-dev \
+    libphp-embed \
+    libargon2-dev \
+    libxml2-dev \
+    libsodium-dev -y
+
+RUN mkdir -p /usr/local/weechat/build
+WORKDIR /usr/local/weechat
+COPY . .
+
+WORKDIR /usr/local/weechat/build
+RUN cmake ..
+RUN make && make install
+ENTRYPOINT ["weechat"]
--- a/README.adoc
+++ b/README.adoc
@@ -112,6 +112,22 @@ They can be launched after compilation from the build directory:
 $ ctest -V
 ----
 
+=== Docker
+
+To optionally build WeeChat with docker:
+
+Build the container:
+
+----
+$ docker build . -t weechat
+----
+
+Run the container to launch WeeChat:
+
+----
+$ docker run -ti weechat
+----
+
 == Copyright
 
 Copyright (C) 2003-2020 Sébastien Helleu <flashcode@flashtux.org>


Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+    ncurses-dev
+
+RUN wget https://weechat.org/files/src/weechat-2.8.tar.xz \
+    && wget https://weechat.org/files/src/weechat-2.8.tar.xz.asc \
+    && gpg --keyserver hkps://keys.openpgp.org --recv-keys A9AB5AB778FA5C3522FD0378F82F4B16DEC408F8 \
+    # Check that the fingerprint is correct by checking the topic on #weechat channel (chat.freenode.net) 
+    # or by asking a developer.
+    && gpg --verify weechat-2.8.tar.xz.asc weechat-2.8.tar.xz \
+    && tar xvf weechat-2.8.tar.xz \
+    && mkdir -p weechat-2.8/build
```
sim642, line 21: Considering this Dockerfile is in the development repository, shouldn't it use the source code from the cloned git repository, which this Dockerfile resides in instead of taking the last stable release code from weechat.org?

Dockerfile
```diff
@@ -0,0 +1,34 @@
...
+
+WORKDIR /weechat-2.8/build
+RUN cmake .. -DENABLE_PHP=OFF \
+    -DENABLE_JAVASCRIPT=OFF \
+    -DENABLE_PYTHON=OFF \
+    -DENABLE_RUBY=OFF \
+    -DENABLE_LUA=OFF \
+    -DENABLE_PERL=OFF \
+    -DENABLE_TCL=OFF \
+    -DENABLE_GUILE=OFF
```
sim642, line 31: >  Some script dependencies are difficult to add to Alpine such as V8

Javascript plugin (based on V8) is problematic in general (issue #360) and also disabled in our Debian builds, if I remember correctly. Other plugins should be fine though and likely should be enabled. Especially Python and Perl plugins, which are required for using popular scripts. Users of a Docker container, who want easy and reproducible WeeChat setup, would likely want to also use scripts as easily.

flashcode: As @sim642 said, I think all plugins except Javascript must be enabled (not only Python and Perl).
If debian base is better for this, then it can be used (I'm more familiar with Debian than Alpine, but I guess Alpine is used here for image size reason only).

Dockerfile
flashcode: If the Dockerfile uses sources from weechat.org, 2 remarks (to be discussed):

1. The file should be `weechat-stable.tar.xz` instead of `weechat-2.8.tar.xz`, there's always a link "stable" to the latest stable release.
2. Like many other projects, this Dockerfile could be moved to a separate repository inside the weechat organisation. Especially if we plan to add multiiple Docker files, for example development version or with different options: for example a container could be setup together with Glowing Bear to provide a graphical interface.

Dockerfile
flashcode: Additional note about Javascript: I'm currently working on disabling this plugin by default, as it compiles only with an old libv8 version, I don't have enough time myself to fix/rewrite the plugin, and got no help past years on this subject (see #360 for more information).

flashcode: Done: JavaScript plugin is now disabled by default: 340d6646a6371e2f224d392c9b1b44c8a20b5074.

flashcode:
By the way, a lot of Dockerfile for WeeChat already exist:

- https://github.com/TheFynx/docker-weechat
- https://hub.docker.com/r/tklx/weechat/
- https://hub.docker.com/r/kerwindena/weechat
- https://hub.docker.com/r/woahbase/alpine-weechat/
- https://hub.docker.com/r/jkaberg/weechat/
- https://hub.docker.com/r/craighurley/docker-weechat/
- https://hub.docker.com/r/dennismp/weechat/
etc.

Dockerfile
yancyribbens: Yes I think it's preferable to use `COPY` to commit the current files from git inside the container if the `Dockerfile` is to committed here.  I was using docker to test and build dependencies such as `gnutls` and so it could be nice to have a dockerfile for that included.  There are advantages to having a dockerfile in a separate repo using `weechat-stable.tar.xz` like being able to add to dockerhub and testing with other integrations (like glowing bear).

Dockerfile
yancyribbens: I'm probably more familiar with Debian myself also, but Alpine does have some nice security features and smaller image size.  Now that JS is disabled, I could update this to just use the default `cmake` with either alpine or debian depending on which seems more feasible to support.

Dockerfile
yancyribbens: I've updated this to `COPY` from the current git directory

Dockerfile
yancyribbens: Added all script deps to the build process.  I had some trouble getting Cmake to find `php-dev` so I switched to `debian:stable-slim` which has a fairly small image size as well.

simonvanderveldt:
Thanks for the PR!
What would be needed to get this finished? How do we ping the relevant maintainers for a review?

flashcode:
@simonvanderveldt: for your information I'm currently working on Docker files for WeeChat (one based on Alpine, one based on Debian), and I'll probably commit them in a new external repository (because they can install different WeeChat version, with a user argument).
This is almost finished (since some months, but was paused because not enough free time on my side).
I have to make some tests and push the files.

I'll close this PR when this will be done.

simonvanderveldt:
@flashcode That's great to hear!
If you need some help to do some testing feel free to ping me :)

flashcode:
Hi,

I'm closing this MR as I just added a new repository to build WeeChat container images: https://github.com/weechat/weechat-container

You can build the following images, with Docker or Podman:
- debian
- debian-slim
- alpine
- alpine-slim

Slim version doesn't include some plugins: script, scripts and spell.

Official images for the latest stable version will be pushed soon to Docker hub.

simonvanderveldt:
@flashcode That's great to hear!
What is/will be the organisation/image on Docker hub?

flashcode:
@simonvanderveldt: I created a weechat organization (https://hub.docker.com/u/weechat), so it will probably be available there, I'll ping you once images are uploaded.

Enissay:
> @simonvanderveldt: I created a weechat organization (https://hub.docker.com/u/weechat), so it will probably be available there, I'll ping you once images are uploaded.

Looking forward for this. The docker image I am using is outdated and giving me some issues :-/

simonvanderveldt:
@flashcode Did you manage to make any progress with this? Can I help somehow?

flashcode:
@Enissay, @simonvanderveldt: sorry I was busy with other things (switch to XDG directories and more recently the switch from freenode to libera.chat).
I'll try to push containers in a few days.

By the way, waiting for them, it's extremely easy to build them yourself, as you can see in this README: https://github.com/weechat/weechat-container/blob/main/README.md#build

simonvanderveldt:
OK, no problem :) thanks for the update!

flashcode:
@simonvanderveldt, @Enissay: I pushed images of version 3.1 in the weechat docker organization.
Please let me know if you have problems with these images, as it's the first time I push to the docker hub.

You should be able to run WeeChat with the default image like this (based on Debian Buster):

    docker run -ti weechat/weechat

Or for the Alpine version (smaller image):

    docker run -ti weechat/weechat:latest-alpine

You can use a volume to persist data on your host, see: https://github.com/weechat/weechat-container#run

PR #1525, opened 2020-06-15
strip colon, if present, from ACCOUNT value

jesopo:
I noticed this after updating to a 2.9-dev build that prints `ACCOUNT` lines:

` -- user has identified as :user`

this pull request is a band-aid. the real solution here is to correctly tokenise IRC data as soon as it's read so `:trailing` parameters are just treated as another param in the argv array.


flashcode:
Hi,
Thanks for the PR, I fixed with a slightly different solution.

PR #1538, opened 2020-07-05
irc: add pointer "irc_nick" in hashtable sent to hook focus callback (closes #1535)

### LINKED ISSUE #1535: irc_nick ptr in focus hashtable
## Feature description

It would be useful to access the prefixes from irc_nick when we press a key in cursor mode on the nicklist. I mean prefixes not prefix, prefix is already there and contains only the higher prefix while prefixes contains all the prefix (so if someone is @ and + prefixes will be @+ while prefix will be @). Or just add the prefixes directly instead of a irc_nick ptr if it's easier

sim642:
Closes #1535.

#### TODO when merging
- [ ] Add `irc_nick` under `hook_focus` in plugin API docs with corresponding WeeChat version constraint.


flashcode:
Merged and doc updated, thanks.

PR #1540, opened 2020-07-12
Solaris ncurses

oicontrib:
Build system (autotools) doesn't find ncurses headers on OpenIndiana (SunOS).
We need -lresolv to link weechat on OpenIndiana because of indefined res_init in wee-network.o


flashcode:
Thanks @evelikov, I close this PR.

PR #1552, opened 2020-08-08
core: Allow to read secrets from external programs (closes #141)

### LINKED ISSUE #141: Add option to read passphrase from another program's stdout

Serranya:
This is my go at #141.

There is some stuff still to do that I know of:
- [ ] Tab auto completion does not work for `/secure setcmd | delcmd`. Not sure where to add this yet
- [ ] The screen becomes a jumbled mess when the called program does something with the terminal. Find a way to repaint the screen.
- [ ] Documentation
- [ ] Add to Changelog

There are probably also some TODOs I missed. 

I opened this PR as a draft to discuss if this approach looks good before I spend more time on this.

### DIFF
--- a/src/core/wee-eval.c
+++ b/src/core/wee-eval.c
@@ -1070,6 +1070,28 @@ eval_replace_vars_cb (void *data, const char *text)
         ptr_value = hashtable_get (secure_hashtable_data, text + 9);
         return strdup ((ptr_value) ? ptr_value : "");
     }
+    if (strncmp (text, "sec.cmd.", 8) == 0)
+    {
+        const char *program;
+        char *value, *status;
+        FILE *in;
+
+        status = NULL;
+        if (!(program = hashtable_get (secure_hashtable_cmds, text + 8)))
+        {
+            return strdup("");
+        }
+        if (!(in = popen (program, "r")))
+        {
+            return strdup("");
+        }
+
+        if ((value = alloca(1024))) {
+            status = fgets (value, 1024, in);
+        }
+        pclose (in);
+        return strdup (status ? value : "");
+    }
     config_file_search_with_string (text, NULL, NULL, &ptr_option, NULL);
     if (ptr_option)
     {
--- a/src/core/wee-command.c
+++ b/src/core/wee-command.c
@@ -5578,6 +5578,17 @@ COMMAND_CALLBACK(secure)
         return WEECHAT_RC_OK;
     }
 
+    /* set secured data cmd */
+    if (string_strcasecmp (argv[1], "setcmd") == 0)
+    {
+        COMMAND_MIN_ARGS(4, "setcmd");
+        hashtable_set (secure_hashtable_cmds, argv[2], argv_eol[3]);
+        gui_chat_printf (NULL, _("Secured data cmd \"%s\" set"), argv[2]);
+        command_save_file (secure_config_file);
+        secure_buffer_display ();
+        return WEECHAT_RC_OK;
+    }
+
     /* delete a secured data */
     if (string_strcasecmp (argv[1], "del") == 0)
     {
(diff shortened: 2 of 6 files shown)


src/core/wee-eval.c
```diff
@@ -1070,6 +1070,28 @@ eval_replace_vars_cb (void *data, const char *text)
...
+        const char *program;
+        char *value, *status;
+        FILE *in;
+
+        status = NULL;
+        if (!(program = hashtable_get (secure_hashtable_cmds, text + 8)))
+        {
+            return strdup("");
+        }
+        if (!(in = popen (program, "r")))
```
sim642, line 1084: Instead of using `popen` directly, WeeChat has `hook_process` and `hook_process_hashtable` for executing external processes. They take care of all the intricacies of managing child processes in an asynchronous way and then the child process is executed in a way that shouldn't mess with WeeChat's own terminal.

That does, however, lead to the core issue: eval needs to return the output from executed process immediately (in a blocking way) instead of asynchronously (which is the WeeChat's way of running external processes).
Currently it seems that the process must output something fast enough but that's a massive data race, no? If the process takes a moment to initialize, sync or decrypt some passwords container, it'd already be too late. On the other hand, WeeChat can't just wait for output in the middle of evaluating an expression because it'd block WeeChat completely.

I think the main reason why this has stayed unsolved so far is that there doesn't seem to be an easy solution to these issues.

Serranya: Ahh I see.

I think we have the following options to deal with the problem.

1. Run the process async. The expression will just return "" until the value is ready. The value could be cached to make sure that this does not occur too often. We could even read the values at startup and cache them for as long weechat runs.

2. Block weechat entirely. This would either happen at startup time (And the programm output is just cached) or when the user requests the value.

Im not really sure if blocking is such a big problem. Yes weechat will hang. But this happens in response to a user initiated action. Requesting the password. I don't think the User is really suprised when weechat hangs until the process returns.

sim642: 1. Just returning `""` initially would be a problem because a password may be required to be available immediately on WeeChat startup, e.g. autoconnecting IRC server or relay server.
2. Caching values is also problematic because then we have to do cache invalidation somehow. Users may change the passwords in their password manager.
3. By caching passwords we somewhat duplicate the function of the password manager and decrease security by having copies of passwords also in WeeChat memory. Proper password managers take security very seriously and use special memory protection techniques to prevent leaks through RAM.
4. Moreover, we can't read values at startup ahead of time (even if we ignore point 1.) because we don't know which eval expressions will be executed later in time. And the queried values may also depend on anything else dynamically, so it's an uphill battle to try to do that.
5. Also, if you're suggesting reading values at startup and caching them, you might as well write a script/plugin that (given a list of password names) queries them (asynchronously) from the password manager and sets them in `/secure`. It still has the issue from point 1.
6. Blocking WeeChat is a big problem, user input or not. If WeeChat is blocked waiting for an external process, IRC connections etc can time out (WeeChat isn't responding to `PING`s at that time). Also all `hook_timer`s would majorly mess up. Basically, the entirety of WeeChat has been designed in an asynchronous manner and changing that means changing everything.

flashcode:
Hi,

Thanks for your PR.

I'm closing it because I implemented a way to read the passphrase with an external program, see: https://github.com/weechat/weechat/issues/141#issuecomment-774487254
So the secured data itself is still stored in `sec.conf` and handled by WeeChat itself.

As commended above, reading secured data itself with an external program is hard to implement without major refactoring, it must be asynchronous and not block WeeChat, but in many cases we need the password immediately (example: to connect to an IRC server).

I hope this first step with the passphrase is enough for most users.

PR #1573, opened 2020-10-14
doc: Update hotlist example to match hotlist_prefix

trygveaa:
hotlist_prefix defaults to "H: ", so the example should show the same to
make it easier to understand what it is.


sim642 (approved):

flashcode:
Thanks!

PR #1587, opened 2020-11-13
Add paths used by clangd language server to .gitignore

Chaz6:
In order for the clangd language server to function properly it requires a file called compile_commands.json in the root directory of the project. It also creates a .cache folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -14,11 +14,13 @@
 *.so.0.0.0
 *.1
 
+.cache/*
 ABOUT-NLS
 autom4te*
 build/*
 builddir/*
 compile
+compile_commands.json
 config.guess
 config.h
 config.h.in*


flashcode:
Hi @Chaz6,

Thanks for your PR.

I understand your need to ignore these files, but this seems specific to a tool you are using, and if we add entries for every tool/editor used, the `.gitignore` could become bloated. For now it contains only files left after a build (autotools, cmake or Debian packaging).

There are other ways to locally ignore these files, for example by editing file `.git/info/exclude` (which has same format as the `.gitignore` file).

Please let me know if that's OK for you, in which case we could close this PR.

Chaz6:
Sure that's fine. I thought it might be useful for others who use clangd.

PR #1592, opened 2020-11-24
core: Prevent switching to start of visited buffers when jumping to next

### LINKED ISSUE #1591: Jump to next visited buffer sometimes goes to the start of the visited buffers list
## Bug summary

If you run `/input jump_next_visited_buffer` right after switching to a buffer, weechat changes to the first buffer in the visited buffers list. That is, it wraps around and goes to the buffer you visited the longest ago. Running `/input jump_previously_visited_buffer` does not make it go back to the buffer you were on (it does nothing). However, if you run `jump_previously_visited_buffer` first after switching to a buffer, and then `jump_next_visited_buffer` it works as expected, the next command does not go past the most recently active buffer.

## Steps to reproduce

1. Open a clean weechat: `weechat -t`
2. Add some buffers: `/eval -s /buffer add 2; /buffer add 3; /buffer add 4`
3. Switch to buffer 2, then 3, then 4 by pressing alt-2 alt-3 alt-4
4. Run `/input jump_next_visited_buffer` or press alt->

## Current behavior

It switches to buffer 1.

## Expected behavior

It does nothing.

## Suggested solutions

Make `/input jump_next_visited_buffer` do nothing when you are already on the most recently visited buffer.

---

- WeeChat version: master
- OS, distribution and version: Arch Linux
- Terminal: Kitty
- Terminal multiplexer (screen/tmux/…/none): none

trygveaa:
If you run /input jump_next_visited_buffer right after switching to a
buffer, weechat changes to the first buffer in the visited buffers list.
That is, it wraps around and goes to the buffer you visited the longest
ago. This patch fixes that.

The reason it happens is that when you switch to a buffer (normally,
i.e. in another way than using jump_previously_visited_buffer/
jump_next_visited_buffer) gui_buffers_visited_index is set to -1 (in
gui_buffer_visited_add). This makes gui_buffer_visited_get_index_next
return 0 because it returns gui_buffers_visited_index + 1, which makes
gui_input_jump_next_visited_buffer jump to the first buffer in the list
of visited buffers.

Fixes #1591

### DIFF
--- a/src/gui/gui-buffer.c
+++ b/src/gui/gui-buffer.c
@@ -4257,6 +4257,7 @@ int
 gui_buffer_visited_get_index_next ()
 {
     if ((gui_buffers_visited_count < 2)
+        || (gui_buffers_visited_index < 0)
         || (gui_buffers_visited_index >= gui_buffers_visited_count - 1))
         return -1;
 


trygveaa:
@flashcode: Isn't this a bugfix, not a feature? I can't see any reason the current behavior is intentional, I thought it was pretty confusing and annoying behavior until I figured out what caused it.

flashcode:
Yes, wrong label, sorry.
I'll merge this after release of 3.1 (to prevent any regression the day before the release).

trygveaa:
Great, thanks!

PR #1597, opened 2020-12-06
add -yes to output of /help voice

dotflac:
currently /help voice does not tell you that you need -yes if you are using *. i have changed it to match /help op, /help deop and /help devoice


flashcode:
Thanks.

I just noticed it's missing as well in command `/dehalfop` and the completion of commands `/dehalfop` and `/halfop` is wrong, I'll fix these issues.

PR #1610, opened 2021-01-29
irc: make default chantypes configurable

mvirkkunen:
`!` was removed from the default chantypes in 4a42cda3a5f5b02d8ba9afae129499c26250d334 which broke `!channels` on at least one old server where the ircd doesn't send a `005` to specify the chantypes at all. If channels aren't recognized as channels, any incoming messages on them are shown as private messages which makes the channel a bit hard to use.

This PR adds a configuration option for chantypes which is only used if the server doesn't specify them, so that such servers can manually be fixed on the client side.

If supporting outdated servers is out of scope, feel free to close.


flashcode:
Thanks, merged.

Maybe a validation of the option value must be added, to prevent problems if the user sets something that can break (not compatible with other server options like nick prefixes).

PR #1621, opened 2021-03-20
buflist: Fix wrong pointers being used in hdata_compare

trygveaa:
This used pointer1 and pointer2 which are pointers to the buffers, but
it should use ptr_hotlist1 and ptr_hotlist1 which are pointers to the
hotlists it is trying to compare.

### DIFF
--- a/src/plugins/buflist/buflist.c
+++ b/src/plugins/buflist/buflist.c
@@ -271,7 +271,7 @@ buflist_compare_buffers (void *data, struct t_arraylist *arraylist,
             else
             {
                 rc = weechat_hdata_compare (buflist_hdata_hotlist,
-                                            pointer1, pointer2,
+                                            ptr_hotlist1, ptr_hotlist2,
                                             ptr_field + 8,
                                             case_sensitive);
             }


trygveaa:
Related to this option, it seems a bit limited when you use hotlist. You can set it to `hotlist.creation_time.tv_sec,number,-active` if you want unread buffers sorted by when they became unread, but that will make all read buffers appear before the unread ones (because they have time evaluated to 0). If you want unread buffers first, and sorted by time ascending, there isn't really any way to do that as far as I can see. You can't insert `hotlist` first in the sort value because that is a pointer, not a boolean, so you end up with an arbitrary order.

I guess to solve this we either need to be able to use eval conditions in the value, or a modifier to mark that a variable should be just compared by existing or not existing, or do that for pointers by default, since it seems a bit useless to compare pointer order.

Note: Technically it seems that you can accomplish this by setting e.g. `-hotlist.nonexistent,hotlist.creation_time.tv_sec,number,-active`, but that seems a bit fragile, and at least hard to realize that this is the way you have to do it.

PR #1622, opened 2021-03-27
buflist, irc: Add tls_version bar item

talisein:
This commit adds the TLS version to servers in the buflist, as well as an optional bar item. This allows users to quickly see if they are connecting to a server with an outdated version like 1.0/1.1, and also be happy when they are connected with 1.3.

Screenshot from 2021-03-26 19-16-40

In the screenshot I customized my local weechat to have the inverse video red for cleartext, but in the PR it's just normal red by default. I don't have any relay/proxy set up so I don't know if this would be annoying for that. I also didn't change the default status bar, instead I customized it for myself.


src/plugins/irc/irc-config.c
```diff
@@ -3227,6 +3246,30 @@ irc_config_init ()
         NULL, NULL, NULL,
         &irc_config_change_color_item_nick_modes, NULL, NULL,
         NULL, NULL, NULL);
+    irc_config_color_item_tls_version_max = weechat_config_new_option (
```
flashcode, line 3249: I would rename this variable `irc_config_color_item_tls_version_ok` as it's for a TLS version that is **not** deprecated.

src/plugins/irc/irc-config.c
```diff
@@ -3227,6 +3246,30 @@ irc_config_init ()
...
+        irc_config_file, ptr_section,
+        "item_tls_version_max", "color",
+        N_("color for max supported tls version in bar item \"tls_version\""),
+        NULL, -1, 0, "green", NULL, 0,
+        NULL, NULL, NULL,
+        &irc_config_change_color_item_tls_version, NULL, NULL,
+        NULL, NULL, NULL);
+    irc_config_color_item_tls_version_1_2 = weechat_config_new_option (
```
flashcode, line 3257: I would rename this variable `irc_config_color_item_tls_version_deprecated` and not mention any TLS version, because in future more TLS versions could be marked as deprecated. So we'll not have to rename this option.

src/plugins/irc/irc-config.c
```diff
@@ -3227,6 +3246,30 @@ irc_config_init ()
...
+        irc_config_file, ptr_section,
+        "item_tls_version_1_2", "color",
+        N_("color for TLS1.2 in bar item \"tls_version\""),
+        NULL, -1, 0, "yellow", NULL, 0,
+        NULL, NULL, NULL,
+        &irc_config_change_color_item_tls_version, NULL, NULL,
+        NULL, NULL, NULL);
+    irc_config_color_item_tls_version_other = weechat_config_new_option (
```
flashcode, line 3265: I would rename this variable `irc_config_color_item_tls_version_disabled` to clearly indicate TLS is disabled (and not "other version").

src/plugins/buflist/buflist-bar-item.c
```diff
@@ -596,6 +597,34 @@ buflist_bar_item_buflist_cb (const void *pointer, void *data,
                                    "format_lag", "");
         }
 
+        /* tls protocol */
+        hdata_irc_server = weechat_hdata_get ("irc_server");
```
flashcode, line 601: Wherever possible in the buflist, I try to be agnostic of the protocol used.
As other protocols are supported via scripts (for example Slack), we could try to be compatible with them by not checking IRC data at all and just relying on the local buffer variable you set (in IRC plugin).
The variable should already be set only if you are connected, so you could directly get the variable value for the format here, no?

src/plugins/buflist/buflist-config.c
```diff
@@ -773,6 +776,30 @@ buflist_config_init ()
         NULL, NULL, NULL,
         &buflist_config_change_buflist, NULL, NULL,
         NULL, NULL, NULL);
+    buflist_config_format_tls_version = weechat_config_new_option (
```
flashcode, line 779: Currently, the name used in WeeChat is "ssl" everywhere in options, but I agree migrating to "tls" is more meaningful as we don't use SSL any more.
Renaming the historical options is just annoying for people upgrading to newer versions, as the server options "ssl*" would be renamed and old values would be lost.
So we can keep this anyway, even if not consistent with all options.

src/plugins/irc/irc-bar-item.c
```diff
@@ -519,6 +521,61 @@ irc_bar_item_lag (const void *pointer, void *data,
...
+irc_bar_item_tls_version (const void *pointer, void *data,
+                          struct t_gui_bar_item *item,
+                          struct t_gui_window *window, struct t_gui_buffer *buffer,
+                          struct t_hashtable *extra_info)
+{
+    char buf[128];
+    struct t_irc_server *server;
+    gnutls_protocol_t version;
+    const char *color;
+    /* make C compiler happy */
```
flashcode, line 538: Please add a blank line before this one.

flashcode (changes requested):
Thanks for the PR!

Additional changes required (see my other comments as well):

- the help on `/buflist` command must be updated with your new format
- the user's guide must be updated: the new bar item must be added in chapter "Screen layout", table below "Other items available (not used in bars by default)" (only the English version, translations will come later, some of them by other contributors).

src/plugins/buflist/buflist-config.c
```diff
@@ -773,6 +776,30 @@ buflist_config_init ()
...
+        NULL, NULL, NULL,
+        &buflist_config_change_buflist, NULL, NULL,
+        NULL, NULL, NULL);
+    buflist_config_format_cleartext = weechat_config_new_option (
+        buflist_config_file, ptr_section,
+        "format_cleartext", "string",
+        N_("format for indicating an unencrypted connection on an IRC "
+           "server buffer (note: content is evaluated, see /help buflist)"),
+        NULL, 0, 0,
+        " ${color:default}(${color:red}cleartext${color:default})",
```
flashcode, line 798: This string must be translated (for the word `cleartext`).
But that's not ideal to have the whole format translated, as the default value would change according to your locale.
I'll think about a better way to translate this.

src/plugins/buflist/buflist-config.c
```diff
@@ -645,7 +648,7 @@ buflist_config_init ()
            "[${number}]}\""),
         NULL, 0, 0,
         "${format_number}${indent}${format_nick_prefix}${color_hotlist}"
-        "${format_name}",
+        "${format_name}${format_tls_version}",
```
flashcode, line 651: I wonder if this should be displayed by default. I think most people doesn't want this by default, but it's hard to know what people prefer.
Personally, I think it's better to not have this by default, to keep a simple format by default.

weechatter: I agree, this should not be a default value.

src/plugins/irc/irc-config.c
talisein: I went with `irc_config_color_item_tls_version_insecure` as it will be used for both cleartext and insecure protocols like TLS1.0

src/plugins/buflist/buflist-bar-item.c
talisein: Done--I added a cleartext annotation to dcc-chat as well as a proof of concept.

src/plugins/buflist/buflist-config.c
talisein: I've created a buflist_config_format_tls_format_default buffer that I print a gettext("cleartext") into. This does make the default translation dependent which is kind of weird, but if you think up a better approach I am happy to change it.

src/plugins/buflist/buflist-config.c
talisein: Removed. Weechat is known for its good defaults so I think you have better taste here.

doc/de/weechat_user.de.adoc
```diff
@@ -839,6 +839,8 @@ andere Items die zur Verfügung stehen (die aber nicht standardmäßig in einer
 | spell_dict                   | `de,en`                  | zeigt an welche Wörterbücher für die Rechtschreibung im aktuellen Buffer genutzt werden.
 | spell_suggest                | `Glück,Glocke,Block`     | Vorschläge für ein falsch geschriebenes Wort.
 | window_number                | `2`                      | Nummer des aktuellen Fensters.
+// TRANSLATION MISSING
+| tls_version                  | `TLS1.3`                 | TLS version in use for current IRC server.
```
flashcode, line 843: Items should be sorted alphabetically in this table (same remark for other languages).

src/plugins/buflist/buflist-bar-item.c
```diff
@@ -596,6 +596,21 @@ buflist_bar_item_buflist_cb (const void *pointer, void *data,
...
+            weechat_hashtable_set (
+                buflist_hashtable_extra_vars,
+                "format_tls_version",
+                weechat_config_string (buflist_config_format_tls_version));
+        }
+        else
+        {
+                weechat_hashtable_set (buflist_hashtable_extra_vars,
+                                       "format_tls_version", "");
+        }
```
flashcode, line 612:
```suggestion
        weechat_hashtable_set (
            buflist_hashtable_extra_vars,
            "format_tls_version",
            (ptr_tls_version && ptr_tls_version[0]) ?
            weechat_config_string (buflist_config_format_tls_version) : "");
```

src/plugins/irc/irc-bar-item.c
```diff
@@ -519,6 +521,65 @@ irc_bar_item_lag (const void *pointer, void *data,
     return NULL;
 }
 
+/*
+ * Returns content of bar item "tls_version": bar item with tls version value.
```
flashcode, line 525:
```suggestion
 * Returns content of bar item "tls_version": bar item with TLS version value.
```

src/plugins/irc/irc-config.c
```diff
@@ -3227,6 +3246,30 @@ irc_config_init ()
         NULL, NULL, NULL,
         &irc_config_change_color_item_nick_modes, NULL, NULL,
         NULL, NULL, NULL);
+    irc_config_color_item_tls_version_ok = weechat_config_new_option (
+        irc_config_file, ptr_section,
+        "item_tls_version_ok", "color",
+        N_("color for max supported tls version in bar item \"tls_version\""),
```
flashcode, line 3252:
```suggestion
        N_("color for max supported TLS version in bar item \"tls_version\""),
```

src/plugins/irc/irc-config.c
```diff
@@ -3227,6 +3246,30 @@ irc_config_init ()
...
+        NULL, -1, 0, "green", NULL, 0,
+        NULL, NULL, NULL,
+        &irc_config_change_color_item_tls_version, NULL, NULL,
+        NULL, NULL, NULL);
+    irc_config_color_item_tls_version_deprecated = weechat_config_new_option (
+        irc_config_file, ptr_section,
+        "item_tls_version_deprecated", "color",
+        N_("color for deprecated tls versions in bar item \"tls_version\""),
```
flashcode, line 3260:
```suggestion
        N_("color for deprecated TLS versions in bar item \"tls_version\""),
```

src/plugins/irc/irc-config.c
```diff
@@ -3227,6 +3246,30 @@ irc_config_init ()
...
+        NULL, -1, 0, "yellow", NULL, 0,
+        NULL, NULL, NULL,
+        &irc_config_change_color_item_tls_version, NULL, NULL,
+        NULL, NULL, NULL);
+    irc_config_color_item_tls_version_insecure = weechat_config_new_option (
+        irc_config_file, ptr_section,
+        "item_tls_version_insecure", "color",
+        N_("color for insecure tls versions in bar item \"tls_version\""),
```
flashcode, line 3268:
```suggestion
        N_("color for insecure TLS versions in bar item \"tls_version\""),
```

flashcode (changes requested):

doc/de/weechat_user.de.adoc
talisein: Fixed, thanks for taking another look.

flashcode (approved):

flashcode:
Merged, thanks!

I made some changes:

- The text "cleartext" is not translated any more in the local variable, so that scripts can rely on this exact text. To do that I added an evaluated form `${translate:xxx}` to translate the text `xxx` into the user language, which is used in the new buflist format.
- Minor changes in translations and order of IRC config options.

PR #1625, opened 2021-04-01
Track registration status in irc server hdata

talisein:
I recently realized I had lost my nick registration for some time when my cert expired. So I'd like an easy way to track
(1) if I've been authenticated into an irc server,
(2) if I used SASL, I want to know which method was used.

So I've added this to the irc server hdata. I'm using it like this, but no default formats are changed. It lets me see where I'm using certfp vs SASL PLAIN vs just normal nickserv identify, and lets me remember to check the PLAIN servers every once in a while if they support certfp yet. And if login fails, I'll get a red marker that makes it obvious.
Screenshot from 2021-03-31 13-53-21

Summary of change:
If SASL login is successful, authentication_method is set to SASL and the sasl_mechanism_used is set to the sasl method used. 

Otherwise, whenever mode +r is set, authentication_method is set to OTHER. If a server uses another mode than 'r' for this, it is configurable. There's no protocol number for getting logged in from PASS or nickserv identify, so I think reacting to the mode is the only option.

### DIFF
--- a/src/plugins/irc/irc-server.h
+++ b/src/plugins/irc/irc-server.h
@@ -94,6 +94,7 @@ enum t_irc_server_option
     IRC_SERVER_OPTION_SPLIT_MSG_MAX_LENGTH, /* max length of messages        */
     IRC_SERVER_OPTION_CHARSET_MESSAGE,      /* what to decode/encode in msg  */
     IRC_SERVER_OPTION_DEFAULT_CHANTYPES,    /* chantypes if not received     */
+    IRC_SERVER_OPTION_REGISTERED_MODE,      /* mode set on registered user   */
     /* number of server options */
     IRC_SERVER_NUM_OPTIONS,
 };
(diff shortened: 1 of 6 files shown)


src/plugins/irc/irc-server.h
```diff
@@ -248,6 +257,8 @@ struct t_irc_server
     time_t last_user_message;       /* time of last user message (anti flood)*/
     time_t last_away_check;         /* time of last away check on server     */
     time_t last_data_purge;         /* time of last purge (some hashtables)  */
+    int authentication_method;      /* authentication method used to login   */
+    int sasl_mechanism_used;        /* SASL method used at login time        */
```
flashcode, line 261: Hi,

Sorry for the late check of this PR.

Why adding `sasl_mechanism_used` there? You already have the server option `sasl_mechanism` that contains the mechanism used to authenticate in case SASL was used for authentication.

Is it because reading the server option is not so easy, because of value inheritance from `irc.server_default.sasl_mechanism`?
If that's the case, we could consider instead adding something in API (like an info) that could be used from evaluated string to get the inherited value of any option (including IRC server options).

Something like this, considering in this example that `irc.server.libera.sasl_mechanism` == `null` and `irc.server_default.sasl_mechanism` == `plain`:

    /eval -n ${info:option_value,irc.server.libera.sasl_mechanism}
    == [plain]

flashcode (changes requested):

src/plugins/irc/irc-server.h
talisein: I think I wanted to create a record of what was actually used, because you can change the option while you are still connected.

flashcode: OK makes sense.

flashcode (approved):

flashcode:
Merged, thanks!

FYI, I made the following changes:

- fixed some code styles
- added upgrade code to retrieve field values after `/upgrade`
- added call to `irc_mode_registered_mode_change` when option `irc.server_default.registered_mode` is changed (it was already done for options in the server).

PR #1632, opened 2021-04-17
doc: Update weechat.bar.*.condition to use info:term_width

blabber:
In cf93e953b the `weechat.bar.*.condition` examples have been changed to use `${info:term_width}` instead of `${window.win_width}`. The user guide still shows the old example. This commit syncs the user guide with the on-line help.

I stumbled over this while trying to configure a condition for the buflist. I don't know how doc updates are handled with weechat. I could replace the example in all doc/*/weechat_user.*.adoc` files, but I can't change the description of the example in all languages (all I can offer is german).


flashcode:
Thanks!
I'll update the docs in the other languages.

PR #1636, opened 2021-05-07
core: Use mallinfo2() when available

talisein:
mallinfo() is deprecated in glibc 2.33

https://sourceware.org/pipermail/libc-alpha/2021-February/122207.html


flashcode:
Thanks, I'll add the detection of `mallinfo2` in `configure.ac` for autotools as well.

Issue #1641, opened 2021-05-21
Documentation could benefit from more extensive cross-referencing


Oblomov:
The weechat documentation is quite extensive (in fact, I would say that it can be a bit overwhelming for the newcomer), but not particularly easy to navigate (for example, the user guide reads very well as a technical reference manual on paper would read, but finding the right thing or how things connect together requires a good dose of mental effort).

I believe this could be improved with more extensive use of cross-referencing. I'm actually working on a couple of additions to the user guide that will hopefully make some features more discoverable, but some changes I have in mind would benefit from some kind of automation. Some examples:

* every time a command is mentioned in an example, it'd be nice if the command name itself was a link to the reference for that command;
* conversely, it would be nice if the command reference had automatic links to all the examples that make use of the command;
* when the various contexts are explained more in detail, it would be nice to have a link to the corresponding key bindings.

Some of these links could be managed through some kind of sidebar or sidenote, which might require some hacking of the AsciiDoctor output style. (As an example: the `buffers_and_windows` section could have a sidebar with links to the corresponding key bindings and the references for the `/buffer`, `/window` and `/layout` commands).

PR #1642, opened 2021-05-21
doc: ref the layout, buffer and window sections and commands

Oblomov:
Add some references between the Screen layout and Buffers and windows
sections, linking also to the appropriate commands.

This should make it easier to discover the `/layout` command and its
relevance to the windows and buffer management.

(Small contribution to GitHub issue #1641)


flashcode:
Thanks, merged with minor style changes.

PR #1647, opened 2021-05-26
bar.size and bar.size_max should refer to lines

FiXato:
From my understanding `weechat.bar.$bar_name.size_max` and `weechat.bar.$bar_name.size` are actually defined in lines, not characters, yet the help text mentioned *chars*.

Thanks to christophergray for pointing this out on IRC.

### DIFF
--- a/src/gui/gui-bar.c
+++ b/src/gui/gui-bar.c
@@ -1438,7 +1438,7 @@ gui_bar_create_option (const char *bar_name, int index_option, const char *value
             ptr_option = config_file_new_option (
                 weechat_config_file, weechat_config_section_bar,
                 option_name, "integer",
-                N_("bar size in chars (0 = auto size)"),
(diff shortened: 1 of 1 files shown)


flashcode:
Hi @FiXato,

I have two remarks:

* You must not update the weechat.pot manually, it's built from source code, so you must change the string in the source code, then translations are rebuilt.
* The size of bar is actually in "chars". Lines apply only to top/bottom bars. The size of left/right bars is a number of char (width). If we want to change something, we could mention the difference for size between top/bottom and left/right bars.

FiXato:
ahhh, yes, that makes sense. :)

So, perhaps something like "bar size in chars (left/right bars) or lines (top/bottom bars) (0 = auto size)"?

flashcode:
Yes, makes more sense, I let you change that in C source code then?

FiXato:
I hope this addresses the issue better. :)

flashcode:
@FiXato: indeed, thanks.

PR #1652, opened 2021-06-01
Annotate public symbols and hide the rest

evelikov:
Currently we export all the symbols from both the weechat executable and
each of the plugins.

Introduce WEECHAT_PUBLIC macro annotating the respective symbols and
hide everything else.

As a result, we shave around 10% of the size of each binary. Savings
should be even more if LTO is used.

Note: since cmake is the only officially supported build system, I did
not bother with autotools.

Signed-off-by: Emil Velikov <emil.l.velikov@gmail.com>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,13 @@ set(CMAKE_SKIP_RPATH ON)
 set(CMAKE_C_FLAGS   "${CMAKE_C_FLAGS} -fsigned-char -Wall -Wextra -Werror-implicit-function-declaration")
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsigned-char -Wall -Wextra")
 
+if(POLICY CMP0063)
+  cmake_policy(SET CMP0063 NEW)
+endif()
+set(CMAKE_CXX_VISIBILITY_PRESET hidden)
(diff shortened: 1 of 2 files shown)


evelikov:
Hmm just realised that this will cause problems with the weechat backtrace functionality. In particular:
The existing code relies on `backtrace(3)` which in its essence seems to work by _mandating_ all the symbols to be exported. On the other hand a far more elegant solutions is to use `libunwind(3)`.

This ancient [Xorg commit](https://gitlab.freedesktop.org/xorg/xserver/-/commit/e21e183059df5975e7086850d1931edb2c1bbd06
) demonstrates the amount of work and most importantly quality of the backtrace across the two. I will work on adding libunwind support as a prep step.

evelikov:
For anyone following, this is a bit more involving than expected.

The `rdynamic & backtrace` to `libunwind` port was trivial.

Although the tests are failing since they use internal API:
 - the weechat one is indirectly exposed via the plugin API, yet the test is not registering an API at that point
 - plugin implementation details

Fixing that results in more convoluted circular dependencies... we already have some. That can be trivially resolved with OBJECT libraries, which requires cmake 3.12 yet it is missing on older Debian/Ubuntu. More elaborate fix of the circular deps requires some reshuffle on the headless/curses side...

So I'm slowly shaving away this lovely yak :grin:

PR #1655, opened 2021-06-03
Initial serbian translation.

eevan78:
Hello,
I'd like to contribute a serbian translation. It is complete translation.

New Cmake files are created or existing ones edited, docgen.py updated
to add 'sr_RS'.
Also added --previous option to msgmerge arguments so that updating of 
fuzzy strings becomes easier with diff. But maybe others don't like that, so I
could remove it from this commit.
There are no changes to debian packaging.


flashcode:
Thanks for this amazing work, merged!

I made minor changes:

* fixed autotools build
* replaced some ^M by UNIX newlines
* removed some trailing whitespace
* removed the `--previous` flag you added (I'm not sure to understand exactly what it does, could you give me an example?)

One thing you could add in docs: a "Translator:" line with your name on the next line, like it's done for example in German docs.

I'll add Serbian language on weechat.org site very soon.

eevan78:
Thanks for fixes, I will update the docs with "Translator:" line.

Regarding the `--previous` flag in msgmerge.
When you change the original English text (especially in msgid with a lot of text) sometimes it is difficult to see what exactly is changed. It could be obvious for a short string. This flag adds an old (previous) msgid as a comment lines of the new msgid, so it is possible to compare new and old msgid to identify changes. Here is an example created from Vim documentation that I translate using po4a:

``` po
#. type: Plain text
#: d:/Projects/doc-prevod/original/TXT/gui.txt:175
#, fuzzy, no-wrap
#| msgid ""
#| "There are vertical scrollbars and a horizontal scrollbar.  You may\n"
#| "configure which ones appear with the 'guioptions' option.\n"
msgid ""
"There are vertical scrollbars and a horizontal scrollbar.  You may\n"
"configure if you like which ones appear with the 'guioptions' option.\n"
msgstr ""
"Постоје вертикалне скрол линије и хоризонтална скрол линија. Опцијом\n"
"'guioptions' бирате које ће се приказивати.\n"
```
Here I have changed the text in source file by adding "if you like".
`--previous` flag added the lines marked with #| that show the previous, old msgid. Now I can use `diff` and easily see what was changed in a new msgid. This is a short example, but feature is much more useful in long strings where you have to carefully inspect the new string and change your translation. It is much faster and easier with this feature.

flashcode:
Nice option, I'll enable it then, thanks.

PR #1657, opened 2021-06-04
irc: allow irc_raw_in/irc_in hooks to eat messages

emersion:
This is useful to implement IRC protocol extensions which introduce
new commands.


flashcode:
Thanks, merged!

PR #1664, opened 2021-06-20
core: Update Serbian translation (sr.po)

eevan78:


flashcode:
Thanks, squashed and merged.

PR #1684, opened 2021-08-11
buflist: Do not hardcode colors in tls_version

KaiserBarbarossa:
Do not hardcode the colors in buflist.format.tls_version

### DIFF
--- a/src/plugins/buflist/buflist-config.c
+++ b/src/plugins/buflist/buflist-config.c
@@ -780,8 +780,10 @@ buflist_config_init ()
         N_("format for TLS version on an IRC server buffer "
            "(note: content is evaluated, see /help buflist)"),
         NULL, 0, 0,
-        " ${color:default}(${if:${tls_version}==TLS1.3?${color:green}:"
-        "${if:${tls_version}==TLS1.2?${color:yellow}:${color:red}}}"
+        " ${color:default}(${if:${tls_version}==TLS1.3?${color:"
+        "${irc.color.item_tls_version_ok}}:${if:${tls_version}==TLS1.2?"
+        "${color:${irc.color.item_tls_version_deprecated}}:"
+        "${color:${irc.color.item_tls_version_insecure}}}}"
         "${translate:${tls_version}}${color:default})",
         NULL, 0,
         NULL, NULL, NULL,


ksu-0:
This incorrectly marks TLS 1.2 as deprecated. At the moment, deprecated versions are TLS 1.1 and earlier, and DTLS 1.0.

KaiserBarbarossa:
> This incorrectly marks TLS 1.2 as deprecated.

That’s what is done at the moment too and which was all introduced in #1622 (Suggested by [this comment](https://github.com/weechat/weechat/pull/1622#discussion_r602684074)). As you can see here, too:

https://github.com/weechat/weechat/blob/ad5fa7c99faf8a1c0b6b9b1f7f6e4b42a03a5c2c/src/plugins/irc/irc-bar-item.c#L562-L564

flashcode:
@ksu-0: where can we check which versions of TLS are deprecated or not (today)?

ksu-0:
The authoritative source for SSL 3.0 and newer, and for DTLS is the status of RFC that defines it. Moreover, each deprecation (including one for the never standardized SSL 2.0) was announced by RFC.
| version | RFC | status | deprecation RFC |
| :-: | :-: | - | :-: |
| TLS 1.3 | [8446](https://www.rfc-editor.org/info/rfc8446) | proposed standard | — |
| TLS 1.2 | [5246](https://www.rfc-editor.org/info/rfc5246) | proposed standard | — |
| TLS 1.1 | [4346](https://www.rfc-editor.org/info/rfc4346) | historic (changed from proposed standard) | [8996](https://www.rfc-editor.org/info/rfc8996) |
| TLS 1.0 | [2246](https://www.rfc-editor.org/info/rfc2246) | historic (changed from proposed standard) | [8996](https://www.rfc-editor.org/info/rfc8996) |
| SSL 3.0 | [6101](https://www.rfc-editor.org/info/rfc6101) | historic | [7568](https://www.rfc-editor.org/info/rfc7568) |
| SSL 2.0 | — | expired draft | [6176](https://www.rfc-editor.org/info/rfc6176) |
| | | |
| DTLS 1.2 | [6347](https://www.rfc-editor.org/info/rfc6347) | proposed standard | — |
| DTLS 1.0 | [4347](https://www.rfc-editor.org/info/rfc4347) | historic (changed from proposed standard) | [8996](https://www.rfc-editor.org/info/rfc8996) |

talisein:
My choice to use the word "deprecated" in the color option was imprecise. That color option is meant to draw attention to servers that are not using the latest TLS version but are still secure. The versions currently officially deprecated are denoted by the "insecure" color option.
 
I think this PR is good. Adjusting the color option name, if needed, can be done separately (this deprecated word is used elsewhere throughout the code for TLS 1.2).

KaiserBarbarossa:
What about renaming "deprecated" to "elder"?

PR #1686, opened 2021-08-22
Minor cleanup

a1346054:
Mainly making sure the license file is identical to what is distributed by GNU, since it says to distribute verbatim.


flashcode:
Thanks!

PR #1692, opened 2021-09-14
core: Update Serbian translation

eevan78:
Update of _sr.po_ and Serbian translation of _Plugin API documentation_

PR #1711, opened 2021-10-15
core: Update Serbian translation

eevan78:
This PR is replacing the #1692. 
I have rebased and squashed all comits into single one.
Hopefully now it can be merged in.


flashcode:
Thanks!

PR #1715, opened 2021-10-26
core: Update Serbian translation

eevan78:
Here is an update of Serbian translation.


flashcode:
Thanks, I've upported your changes on the master branch and committed your changes.

PR #1720, opened 2021-11-08
Update Turkish translations

bitigchi:
This is another attempt to update translations after rebasing on top of master it and starting afresh. Hopefully this will pass the CI tests.


bitigchi:
Regarding the `gettext` errors in CI builds, I believe it's not fair to call them errors, since all of them are intentional. For Turkish, due to completely opposite word order (compared to English), some translations simply sounds weird without a colon to describe the following variable, or some strings simply contain more lines due to the translation being long or intentionally formatting them to fit in 80-column terminals. Wouldn't it be better to alter the tests so that it would only fail at fatal or compile errors?

flashcode:
Hi,

I think most of errors (maybe except 1 or 2) can be fixed: most errors are split of lines that should not be done, even in Turkish, because this will change the way help on commands will be displayed for example.

Split of lines can even cause bugs in the way messages are displayed and handled: a message on two lines will be displayed as 2 separate messages in WeeChat, which is not the case in English. This can affect filters/scripts reacting on such messages.

bitigchi:
Thanks for the information. I've fixed the newlines, only left four columns, which I believe to be essential. It might be possible to fix those with strings containing the property itself, so that I can place the words after the property.

bitigchi:
Please don't merge this yet, as I've been translating and checking a lot recently; I'll push another commit before the merge.

bitigchi:
Alright, the translations have been completed. Aside from three intentional colons for explanatory purposes, it passes all `msgcheck`s. Spell checking has been performed as well. Good to go!

flashcode:
Hi,

Why intentional colons added? Is it mandatory because of Turkish rules or did you add them because you think they are missing in English as well? (in which case you should probably remove them)

bitigchi:
Hello,

Turkish has a different word order (subject + object + verb), therefore in some cases it sounds ambiguous. It's not a problem if the following variable is included in the string (`verb %s`), since it can be rearranged, but if not (`verb` + variable) it is best to use a colon to indicate the subject or object.

flashcode:
So I'll add "noqa" marker on these lines, so that msgcheck will be happy and CI will be green.

flashcode:
Merged.

bitigchi:
Thank you! Happy to have weechat fully translated.

PR #1724, opened 2021-12-01
relay: Add "sync * hotlist" option

nilscc:
Hi,

this PR adds an option for a relay client to sync on the "hotlist_changed" hook signal.

Please review and let me know if anything should be changed.

TODO: Update documentation.


src/plugins/relay/weechat/relay-weechat-protocol.c
```diff
@@ -840,6 +843,68 @@ RELAY_WEECHAT_PROTOCOL_CALLBACK(completion)
...
+
+        /* send signal only if sync with flag "hotlist" */
+        if (relay_weechat_protocol_is_sync (ptr_client, NULL,
+                                            RELAY_WEECHAT_PROTOCOL_SYNC_HOTLIST))
+        {
+            msg = relay_weechat_msg_new (str_signal);
+            if (msg)
+            {
+                snprintf (cmd_hdata, sizeof (cmd_hdata),
+                          "hotlist:0x%p", ptr_hotlist);
```
flashcode, line 895: Even if this is equivalent, please use `0x%lx` here, as done in the other functions.

flashcode (changes requested):
Apart from the small point noted in the comment, seems OK for me!
I'll update the docs in a separate commit.

src/plugins/relay/weechat/relay-weechat-protocol.c
nilscc: Done. I also fixed the creation_time members and removed the next/prev hot list pointers, as they are always null anyway.

flashcode (approved):

nilscc:
PR converted to draft as I'm observing weechat crashes with these changes. I'm investigating.

2xsaiko:
Would this allow for sending the hotlist state from the relay to the clients, so kind of the reverse of what is already there with clients sending their read status?

nilscc:
@2xsaiko yeah, sure. it saves you from syncing on all buffers while still giving you notice of highlights etc.

nilscc:
@flashcode What would be missing to get this change merged?

flashcode:
We have to check if the add of line hdata could be problematic for existing clients. It should not, but it depends how the clients implement the protocol, it can break.

Once we validated the major relay clients are OK with this change, this can be merged (I'll add appropriate documentation as well).

Did you test some relay clients with these changes?

nilscc:
Closing this PR as I personally will be focusing on the new API.

PR #1725, opened 2021-12-01
Update README for optional dependencies.

xloem:
Many of the optional dependencies are enabled by default for cmake, and it isn't mentioned how to disable them.

Autotools actually seems a little more maintained in this regard, where, for me and unlike cmake, it will automatically disable some features requiring missing optional dependencies.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -68,7 +68,9 @@ WeeChat can be built with https://cmake.org/[CMake] (recommended) or autotools.
 [NOTE]
 Only CMake is officially supported to build WeeChat. You should only use
 autotools if you are not able to use CMake. +
-Building with autotools requires more dependencies and is slower than with CMake.
+Building with autotools is slower than with CMake.
+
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,

Thanks for your PR.

Using autotools require mode dependencies, but not the ones mentioned in the user's guide: some packages required by autotools stuff.

Anyway I think I should completely remove the "Install" chapter from README as it's a duplicate of the user's guide, and instead provide a link to the user's guide there.
I'm not fan of having duplicated code/doc.

What do you think?

xloem:
As a source-building user, it is much easier to have installation instructions in the README.  But as a maintainer, it is always better to keep only one copy of things to update, and it's very normal to have building instructions hidden away since most people don't use them.  cmake and autotools are also very normal.  I'd say go for it.

flashcode:
OK thanks.

Done via this commit: 64592951b5bb2656838531972d03168f78fef262.

PR #1732, opened 2021-12-23
irc: Fix regression causing newlines in MOTD to be stripped

trygveaa:
Commit 9063546ac caused newlines in MOTD to not be printed anymore, this
fixes the regression.


flashcode:
Hi,

Thank you for the fix.
I fixed another way that applies to all numeric messages in this commit: d5c391b1ee57730ed6f94c3d7ee5661ceaff35f5
So I close this PR.

PR #1744, opened 2022-01-18
Remove old and unused config option weechat.plugin.debug

trygveaa:
As far as I can see, this option is not used anywhere and should have
been removed in commit 6e69f7f3c.


flashcode:
Thanks!

PR #1761, opened 2022-03-03
scripts: allow null values in config_new_option

trygveaa:
The plugin API function config_new_option accepts null as the
default_value and/or value however the scripting APIs (except for lua)
didn't allow sending null as a parameter value, so it was impossible to
use it this way. This allows sending a null value for these parameters.

Lua already supported sending in nil for these parameters and it works
as expected, so nothing is changed for this plugin.

For Guile you can now send in #nil, for JavaScript null or undefined,
for Perl undef, for PHP NULL, for Python None, for Ruby nil and for Tcl
$::weechat::WEECHAT_NULL.

In all of these languages except Tcl this is the special value
indicating a missing value. However Tcl only has one type, string, so it
doesn't have a null value. Therefore I created a constant with the value
`\uFFFF\uFFFF\uFFFFWEECHAT_NULL\uFFFF\uFFFF\uFFFF` which is used
instead. This is very unlikely to appear unintentionally. Using the
unicode code point \uFFFF was suggested on
https://wiki.tcl-lang.org/page/null.

I tested this with these scripts:
https://gist.github.com/trygveaa/f91977dde5d2876d502bf55fbf9b50cc

### DIFF
--- a/doc/en/weechat_plugin_api.en.adoc
+++ b/doc/en/weechat_plugin_api.en.adoc
@@ -6445,7 +6445,7 @@ Script (Python):
 # prototype
 def config_new_option(config_file: str, section: str, name: str, type: str, description: str,
                       string_values: str, min: int, max: int,
-                      default_value: str, value: str, null_value_allowed: int,
+                      default_value: str | None, value: str | None, null_value_allowed: int,
                       callback_check_value: str, callback_check_value_data: str,
                       callback_change: str, callback_change_data: str,
                       callback_delete: str, callback_delete_data: str) -> str: ...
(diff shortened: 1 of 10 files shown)


ksu-0:
I would suggest to make part of the Tcl null constant random. It would prevent the value to be intentionally fed to Tcl script, assuming that the random part is cryptographically secure, long enough, and not leaked.

A downside of this proposal would be that the null constant, if stored externally, would be not interpreted correctly by another installation.

trygveaa:
> It would prevent the value to be intentionally fed to Tcl script

What's the problem with this, why should it be prevented?

ksu-0:
It can cause null to appear where script author expects non-empty string, which may lead to strange bugs. I can't think of realistic example of the problem happening with `config_new_option`, but I suspect that null, once introduced, may be used again.

trygveaa:
Well, that's the reason I chose a really specific string with non-characters. I don't really see that coming from anywhere else.

ksu-0:
As discussed in the channel, Libera is happy to pass `U+FFFF` through.

flashcode:
Thanks!

PR #1764, opened 2022-03-16
core: Fix regex for constants in the Python stub generator

trygveaa:
In Python raw strings, newlines can't be escaped with a backslash. If
you do that, both the backslash and the newline become part of the
string. This meant that the regex for constants both started and ended
with a newline which caused every other constant to be skipped.


flashcode:
Thanks!

PR #1766, opened 2022-03-21
Update Serbian translations of core and docs

eevan78:
Hi,
I'm sending two commits, one that updates the core translation and one for the update of documentation.


flashcode:
Thanks!

PR #1788, opened 2022-06-29
php: Add stubs and generate arginfo for PHP API

adsr:
As noted in a previous TODO, this was preventing the PHP plugin from working if PHP was compiled with `--enable-debug`. With this patch, `--enable-debug` builds of PHP should work. It also enables use of strict type-checking with the WeeChat API.


flashcode:
Thanks!

PR #1789, opened 2022-06-29
php: Deprecate `forget_*` functions in PHP 8+

adsr:
These functions don't work with PHP 8+ and also don't seem to be necessary anymore when reloading scripts.


flashcode:
Thanks!

PR #1793, opened 2022-07-10
don't require account-notify/away-notify for WHOX

jesopo:
when you connect weechat to znc, you first get a `CAP` dance like this:

```
-> :irc.znc.in CAP unknown-nick LS :batch cap-notify echo-message multi-prefix server-time userhost-in-names znc.in/batch znc.in/self-message znc.in/server-time-iso
<- CAP REQ :cap-notify multi-prefix server-time userhost-in-names
<- CAP END
-> @time=2022-07-10T20:19:26.599Z :irc.znc.in CAP * ACK :cap-notify multi-prefix server-time userhost-in-names
```

and then once znc has figured out what network you're connecting to, you get this:
```
-> @time=2022-07-10T20:19:26.608Z :irc.znc.in CAP user NEW :account-notify away-notify extended-join
<- CAP REQ :account-notify away-notify extended-join
```

and then it takes until after znc has already finished replaying `NAMES` to you for you to get this
```
-> @time=2022-07-10T20:19:26.621Z :irc.znc.in CAP user ACK :account-notify away-notify extended-join
```

which means by the time you're hitting the code that I'm changing, you don't know that the network you're on supports `away-notify` or `account-notify`, which means this `if` will reliably fail and weechat simply won't send a WHOX request. this can be solved by simply sending WHOX whenever a network supports it.

### DIFF
--- a/src/plugins/irc/irc-channel.c
+++ b/src/plugins/irc/irc-channel.c
@@ -780,30 +780,18 @@ irc_channel_check_whox (struct t_irc_server *server,
 {
     if ((channel->type == IRC_CHANNEL_TYPE_CHANNEL) && channel->nicks)
     {
-        if (weechat_hashtable_has_key (server->cap_list, "away-notify")
-            || weechat_hashtable_has_key (server->cap_list, "account-notify")
-            || ((IRC_SERVER_OPTION_INTEGER(server, IRC_SERVER_OPTION_AWAY_CHECK) > 0)
-                && ((IRC_SERVER_OPTION_INTEGER(server, IRC_SERVER_OPTION_AWAY_CHECK_MAX_NICKS) == 0)
-                    || (channel->nicks_count <= IRC_SERVER_OPTION_INTEGER(server, IRC_SERVER_OPTION_AWAY_CHECK_MAX_NICKS)))))
+        channel->checking_whox++;
+        if (irc_server_get_isupport_value (server, "WHOX"))
         {
-            channel->checking_whox++;
-            if (irc_server_get_isupport_value (server, "WHOX"))
-            {
-                /* WHOX is supported */
-                irc_server_sendf (server, IRC_SERVER_SEND_OUTQ_PRIO_LOW, NULL,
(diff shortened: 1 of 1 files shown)


jesopo:
an alternative solution is to check what `CAP`s you know the other side has said they support, rather than checking what's actually been acked, but that seems untidy

src/plugins/irc/irc-channel.c
```diff
@@ -780,30 +780,18 @@ irc_channel_check_whox (struct t_irc_server *server,
 {
     if ((channel->type == IRC_CHANNEL_TYPE_CHANNEL) && channel->nicks)
     {
-        if (weechat_hashtable_has_key (server->cap_list, "away-notify")
-            || weechat_hashtable_has_key (server->cap_list, "account-notify")
-            || ((IRC_SERVER_OPTION_INTEGER(server, IRC_SERVER_OPTION_AWAY_CHECK) > 0)
-                && ((IRC_SERVER_OPTION_INTEGER(server, IRC_SERVER_OPTION_AWAY_CHECK_MAX_NICKS) == 0)
```
flashcode, line 786: Hello,
You are removing the use of options away_check and away_check_max_nicks (this one becomes completely unused in IRC plugin), is that wanted?

flashcode (changes requested):

src/plugins/irc/irc-channel.c
jesopo: yes, I don't think there's a reason to ever not use WHOX when it is available

flashcode: But these options are there to let the user enable or disable WHO (or WHOX), based on how many nicks are in the channel. Purpose is to limit the bandwith used (WHO responses can be huge on some channels).

Using WHOX when available is OK for me, but I would like to keep the feature to completely disable WHO(X) if needed.

jesopo: fair enough. i'll tweak the PR in a bit

PR #1799, opened 2022-07-30
Update Turkish translations

bitigchi:


flashcode:
Hi,

Thanks for the updated, now merged.

I noted two things:

1. There's one missing translation:

```po
#, c-format
msgid ""
"%s%s: disconnected from %d server (SSL connection not supported with upgrade)"
msgid_plural ""
"%s%s: disconnected from %d servers (SSL connection not supported with "
"upgrade)"
msgstr[0] ""
```

2. I'm wondering if the plural form is correctly handled in Turkish translations: you declared in the header:

```
Plural-Forms: nplurals=1; plural=0;
```

while the gettext documentation recommends for Turkish (see: https://www.gnu.org/software/gettext/manual/html_node/Plural-forms.html#Plural-forms):

```
Plural-Forms: nplurals=2; plural=n != 1;
```

This results as this translation in your file:

```po
#, c-format
msgid "%d certificate loaded (system: %d, user: %d)"
msgid_plural "%d certificates loaded (system: %d, user: %d)"
msgstr[0] "%d sertifika yüklendi (sistem: %d, kullanıcı: %d)"
```

While we could expect something like this, with two translations (singular / plural, `msgstr[1]` must be updated accordingly):

```po
#, c-format
msgid "%d certificate loaded (system: %d, user: %d)"
msgid_plural "%d certificates loaded (system: %d, user: %d)"
msgstr[0] "%d sertifika yüklendi (sistem: %d, kullanıcı: %d)"
msgstr[1] "%d sertifika yüklendi (sistem: %d, kullanıcı: %d)"
```

bitigchi:
You're right, looks like the header was initially set wrong. I'll correct it, and re-update the strings file.

PR #1812, opened 2022-08-18
Update alias-config.c to add /GHOST

saberan0n:
Hi, this update would add /GHOST $1 $2 as an alias to /msg NickServ GHOST <nickname> <password> by default


flashcode:
Hello,

Thanks for your pull request.

I'm not really against adding such alias, but this could lead to the add of all nickserv/chanserv sub-commands, which are not something standard.

I think it's better for each user to add its own aliases instead of making them default in config.

Is that OK for you?

PR #1823, opened 2022-09-28
doc/scripting: Add missing WEECHAT_HOOK_PROCESS_CHILD constant

trygveaa:
This constant for the hook_process method was missing from the scripting
docs and thus also from the Python typings.


flashcode:
Hi,

This constant is used only in C API, not scripting API, see the note where it's defined: https://github.com/weechat/weechat/blob/master/src/plugins/weechat-plugin.h#L157

So I close this PR.

PR #1824, opened 2022-09-28
doc/api: Add values for hook_process constants

trygveaa:
When logging this value I just see a number so I have to look up what it means. Previously you would have to check the code or print the value of each of these constants to see it. Seeing the value directly in the documentation makes this much easier.

### DIFF
--- a/doc/en/weechat_plugin_api.en.adoc
+++ b/doc/en/weechat_plugin_api.en.adoc
@@ -9468,9 +9468,9 @@ Arguments:
 **** _3_: not enough memory
 **** _4_: error with a file
 *** _< 0_:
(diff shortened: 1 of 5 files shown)


flashcode:
Hi,

This constant is used only in C API, not scripting API, see the note where it's defined: https://github.com/weechat/weechat/blob/master/src/plugins/weechat-plugin.h#L157

So if you agree, I'll just close this PR.

trygveaa:
Ah, okay

trygveaa:
@flashcode: Wait, didn't you mean to comment this in PR #1823? This PR is still relevant.

flashcode:
Sorry, I closed the wrong one.

PR #1825, opened 2022-09-28
doc/api: Add types for Python callbacks

trygveaa:


doc/en/weechat_plugin_api.en.adoc
```diff
@@ -15006,13 +15006,13 @@ see example below (supported only in WeeChat ≥ 0.4.2).
...
 # example (callback without "buffer" and "extra_info")
-def my_build_callback(data, item, window):
+def my_build_callback(data: str, item: str, window: str) -> str:
     return "my content"
 
 bar_item = weechat.bar_item_new("myitem", "my_build_callback", "")
 
 # example (callback with all arguments, for WeeChat ≥ 0.4.2)
-def my_build_callback2(data, item, window, buffer, extra_info):
+def my_build_callback2(data: str, item: str, window: str, buffer: str, extra_info: Dict[Any, Any]) -> str:
```
trygveaa, line 15015: Wasn't quite sure what to type `extra_info` to here, since it's always an empty dict, and I'm not allowed to use the `Dict` type without types for the key and value, so I went for `Any` for the key and value.

flashcode: This parameter `extra_info` is not used today (reserved for future use).

Anyway if one day it is used, as all hashtables in scripting API, both key and value will be type `str` so you can set this type.

doc/en/weechat_plugin_api.en.adoc
```diff
@@ -10437,7 +10437,7 @@ def hook_print(buffer: str, tags: str, message: str, strip_colors: int, callback
 
 # example
 def my_print_cb(data: str, buffer: str, date: str, tags: str, displayed: int, highlight: int, prefix: str, message: str) -> int:
-    if int(highlight):
```
flashcode, line 10440: This cast was to be compatible with WeeChat < 1.0 where the `highlight` parameter was string instead of integer (this was a bug), so this makes code compatible with all WeeChat version (> 0.3.0).

But since 1.0 was released in August 2015, I agree we can now remove this.

doc/en/weechat_plugin_api.en.adoc
trygveaa: > This parameter `extra_info` is not used today (reserved for future use).

Yes, I saw, but I noticed that I did get a dict, so wanted to type it as such even though it's always empty.

> Anyway if one day it is used, as all hashtables in scripting API, both key and value will be type `str` so you can set this type.

Alright, changed.

doc/en/weechat_plugin_api.en.adoc
trygveaa: Ah, okay.

I noticed that `remaining_calls` in the `hook_timer` callback is a string even though it's an integer in the C callback. Is there any reason for this, or is that also a bug?

flashcode: I see no reason why this is a string in scripting API only… this was introduced in 2009 and is a bug for me.

Please keep `str` for now, I'll fix in a separate commit once this PR is merged.

flashcode:
Thanks!

PR #1826, opened 2022-09-29
doc/api: Fix config_new_option types for non English docs

trygveaa:
Only English was changed in commit 197a7a01e.

### DIFF
--- a/doc/fr/weechat_plugin_api.fr.adoc
+++ b/doc/fr/weechat_plugin_api.fr.adoc
@@ -6698,7 +6698,7 @@ Script (Python) :
 # prototype
 def config_new_option(config_file: str, section: str, name: str, type: str, description: str,
                       string_values: str, min: int, max: int,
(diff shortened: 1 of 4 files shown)


trygveaa:
I also noticed that this syntax for union types (`A | B`) is [only supported in Python >= 3.10](https://peps.python.org/pep-0604/). If you try to run code with it on prior versions you get a syntax error. Should we rather write it as `Union[A, B]` which is supported in all versions that supports types?

Note that `Union` has to be imported from `typing`, like `Dict`.

flashcode:
Hi,

Thanks for the fix.
I think keeping `A | B` is OK, it's clear as syntax, and I'm not sure people will copy/paste the prototype anyway.

trygveaa:
> I think keeping `A | B` is OK, it's clear as syntax, and I'm not sure people will copy/paste the prototype anyway.

Yeah, right, it will just be read from `weechat.pyi` by type checkers which is just while developing so probably new enough.

PR #1827, opened 2022-10-02
python: Include constant values in python stub

trygveaa:
This is useful for two reasons:

1. When running unit tests for a script weechat needs to be mocked. By having the constant values available in the stub file, they can be loaded from that, instead of having to define the constants manually for the mock.

2. If you log a constant value you have to look up what it means. This makes it easier, in the same vein as PR #1824.


flashcode:
Thanks!

PR #1828, opened 2022-10-02
python: Include script examples in function docstring

trygveaa:
This makes it possible to see how functions are used without having to go to the web page. It's especially useful to see the types of the callback functions.

### DIFF
--- a/src/plugins/python/weechat.pyi
+++ b/src/plugins/python/weechat.pyi
@@ -48,202 +48,436 @@ WEECHAT_HOOK_SIGNAL_POINTER: str
 
 
 def register(name: str, author: str, version: str, license: str, description: str, shutdown_function: str, charset: str) -> int:
-    """`register in WeeChat plugin API reference <https://weechat.org/doc/api/#_register>`_"""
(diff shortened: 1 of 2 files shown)


src/plugins/python/weechat.pyi
```diff
@@ -48,202 +48,454 @@ WEECHAT_HOOK_SIGNAL_POINTER: str
 
 
 def register(name: str, author: str, version: str, license: str, description: str, shutdown_function: str, charset: str) -> int:
-    """`register in WeeChat plugin API reference <https://weechat.org/doc/api/#_register>`_"""
+    """`register in WeeChat plugin API reference <https://weechat.org/doc/api/#_register>`_
+    ::
+        ----
+
+        [NOTE]
```
flashcode, line 55: The parsing of example seems to have failed for this `register` function.

flashcode (changes requested):

src/plugins/python/weechat.pyi
trygveaa: Ah, sorry I missed that. It's fixed now.

PR #1831, opened 2022-10-08
tests: Use LONGS_EQUAL for version number tests

trygveaa:
util_version_number returns an int, but BYTES_EQUAL only compares the last byte of the int, so it wouldn't test the rest of the int.

I see other tests for functions returning an int use LONGS_EQUAL, so that seems correct to use.


flashcode:
Thanks!

PR #1839, opened 2022-10-12
python: Fix return types for config option callbacks

trygveaa:
I erroneously typed the return types for these to int in commit e0c117e14, but they should be None.

### DIFF
--- a/doc/en/weechat_plugin_api.en.adoc
+++ b/doc/en/weechat_plugin_api.en.adoc
@@ -6604,10 +6604,10 @@ def option4_check_value_cb(data: str, option: str, value: str) -> int:
     return 1
     # return 0
 
-def option4_change_cb(data: str, option: str) -> int:
+def option4_change_cb(data: str, option: str) -> None:
     # ...
 
-def option4_delete_cb(data: str, option: str) -> int:
+def option4_delete_cb(data: str, option: str) -> None:
     # ...
 
 option1 = weechat.config_new_option(config_file, section, "option1", "boolean",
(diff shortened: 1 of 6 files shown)


trygveaa:
Uh, no, the Python API actually requires an int returned, even though the C API doesn't want a return value, and the return value in the Python API doesn't seem to be used for anything. Is this a bug in the API?

trygveaa:
I suppose it goes under this in the scripting documentation:

> Almost all WeeChat callbacks must return WEECHAT_RC_OK or WEECHAT_RC_ERROR (exception is modifier callback, which returns a string).

But not really sure what the point is if it isn't used?

flashcode:
Hi,

Even if not used, all callbacks must actually return int, in both C and scripting API.

The return value, if not used today, may be in future.

So for me the docs are good as-is.

flashcode:
Ah, you're right.
These 2 callbacks are somewhat special and return void.
So this must be the same behavior in API, ie we must ignore the return value and not expect an integer.

trygveaa:
> Ah, you're right. These 2 callbacks are somewhat special and return void. So this must be the same behavior in API, ie we must ignore the return value and not expect an integer.

You haven't done this yet? I don't think this PR should be merged before that is done, since the callbacks require you to return an int now.

flashcode:
Right, not done yet, and I merged this PR too early.
Not a major issue, I'll fix the scripting API tomorrow.

trygveaa:
Alright, great!

flashcode:
Fixed in 56dbd871ea5fb4a1aa27f74eba8907f0a962db7f (and tests on scripting API config options added before as well: a1057c9a035e77db257158af98153ae69882e19e).

trygveaa:
Nice!

PR #1843, opened 2022-10-22
scripts: Send null values to config section callbacks

trygveaa:
The callback_read and callback_create_option functions in the scripting
APIs always get the value as a string, never as null. This means that if
the value is null, there is no way for the script to distinguish this
from an empty string for string options. This makes it impossible to
properly make options with fallback values, like the irc server and
server_default options, as far as I can see.

All the scripting languages except Tcl use that language's equivalent
for null. For JavaScript which has both null and undefined, null is
used. For Tcl, the magic null string defined in commit 197a7a01e is used
and the documentation is updated to describe that.

I tested this with these scripts:
https://gist.github.com/trygveaa/2d49c609addf9773d2ed16e15d1e3447

You can load all of those scripts and see the result with this command
(assuming you have the scripts in the current directory):

    weechat -t -r "/filter add script * * script; /script load $(echo script_config.*)"

### DIFF
--- a/doc/de/weechat_scripting.de.adoc
+++ b/doc/de/weechat_scripting.de.adoc
@@ -201,11 +201,22 @@ Funktionen werden aufgerufen mittels `+weechat.xxx(arg1, arg2, ...)+`.
 
 Funktionen werden aufgerufen mittels `+weechat::xxx arg1 arg2 ...+`.
 
-Da Tcl nur String-Typen hat, gibt es keinen Null-Typ, der als Argument übergeben werden kann
-wenn eine Funktion Nullwerte akzeptiert. Um dies zu überwinden, können Sie die Konstante
-`$::weechat::WEECHAT_NULL` verwenden, das als Nullwert fungiert. Diese Konstante ist definiert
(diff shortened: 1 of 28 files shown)


flashcode:
This change is OK for me.

I think for TCL the special NULL value should be sent to the callback, and this should be mentioned as something specific in the scripting guide, there: https://weechat.org/files/doc/devel/weechat_scripting.en.html#language_tcl

Documentation in other languages should be updated as well.

trygveaa:
This is now updated for all languages, and I updated the commit message and PR description to reflect that. I also added a link to scripts I wrote to confirm that it works.

I used the magic value for Tcl and have updated the documentation. I copied the English text over to the other languages and added `TRANSLATION MISSING`. For the ones that were already translated, I replaced the translations since I changed the text. Please tell me if that was wrong.

Note that I initially added a new format specifier, `z` for string or null, but I realized that it just makes it more inconvenient than to just use `s` for string or null. If we had `z` for string or null and `s` for non-null string, we would have to raise/print an error if null was sent into `s` by mistake, or it would crash. I don't see any benefit of separating these cases, so I just changed `s` to support null. It turns out this is how it is in the Python C API too, `s` and `z` is the same, so I removed the change to use `z` that I had in the initial version here.

PR #1865, opened 2022-12-25
ruby: add detection of Ruby 3.2

dawidd6:
Ruby 3.2 has been released, it is currently not possible to build weechat with it and this PR aims to change that.

based on https://github.com/weechat/weechat/commit/d0c857934b1ce857823b63031439eb00663cada8


flashcode:
Thanks!

PR #1885, opened 2023-02-21
Add line pointer to line_data struct

nilscc:
Hi!

I would like to add a pointer to the `t_gui_line` struct to the `t_gui_line_data` struct, mainly to reference the next/previous line easily from a `_buffer_line_added` relay message.

### DIFF
--- a/src/gui/gui-line.c
+++ b/src/gui/gui-line.c
@@ -1299,6 +1299,7 @@ gui_line_mixed_add (struct t_gui_lines *lines,
     new_line = malloc (sizeof (*new_line));
     if (new_line)
     {
+        line_data->line = new_line;
         new_line->data = line_data;
         gui_line_add_to_list (lines, new_line);
     }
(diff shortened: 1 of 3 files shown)


flashcode:
Hi,

Thanks for the PR.

Some remarks:

1. Multiple lines can point to same line_data structure (that's why there's a separate structure there), when buffers are merged.
2. Adding a pointer in the line_data structure will use some extra memory.
3. Relay has already the line pointer, but it's not given to the client, we could try to find a way to give it, while staying compatible with existing clients.

I think we need a different approach to send the line pointer to the relay clients.

nilscc:
Current version sends two hdata objects on `_buffer_line_added`, one for the `line_data` and one for the original `line` sent in the core hook.

Since two instead of one objects are sent relay clients with this new logic should either:

 * check the hPath if the hdata object is `line` or `line_data` (most stable) or
 * only use the first hdata object (less stable if order is ever changed)

Maybe this could be added as hint to the documentation if this current approach is taken.

nilscc:
Closing this PR as I will be focusing on the API instead.

PR #1888, opened 2023-03-07
core: Allow /eval to get hashtable properties

talisein:
Some hashtables have empty values, so in an /eval context there is no difference between a key that exists with no value, and a key that doesn't exist. One way to surface this information is to make the `weechat_hashtable_get_string()` properties available, as done in this PR.

Example:
/eval -n ${irc_server[irc_servers].cap_ls.keys()}
[away-notify,chghost,invite-notify,multi-prefix,sasl,userhost-in-names]


flashcode:
Thanks, squashed and merged!

PR #1891, opened 2023-03-13
core: Perform hdata_search in /eval

talisein:
Without search, hdata lists are cumbersome to access in triggers. This PR introduces a search operator:
`hdata[list?search_condition].var1.var2...`
The search condition can contain subexpressions, but instead of bracketed `${foo}`, a parenthesized `$(foo)` must be used. For example:

Fetch the description of a loaded plugin with name == "irc"
```
>> ${plugin[weechat_plugins?$(plugin.name) == irc].description}
== [IRC (Internet Relay Chat) protocol]
```

Fetch the irc server capability string for the server who's name matches the current buffer's name split on periods:
```
>> ${irc_server[irc_servers?$(irc_server.name) == $(split:1,.,,$(buffer.name))].cap_ls.keys()}
== account-notify,away-notify,chghost,extended-join,multi-prefix,sasl,tls,account-tag,cap-notify,echo-message,server-time,solanum.chat/identify-msg,solanum.chat/oper,solanum.chat/realhost
```


flashcode:
Hi,

Thanks for this PR.

I think it's not a good idea to use another prefix/suffix for subexpressions when searching in hdata, for 2 reasons:

- the prefix and suffix can be customized in eval: not directly with `/eval` but when using API function `string_eval_expression`, so for example if you choose `$(` as prefix and `)` as suffix, this will conflict with the ones you just added for hdata search ;
- it's confusing for the user to know when to use one prefix or the other one.

Now the problem is how to have an evaluated condition that is not immediately evaluated.

Something similar is done with the format `${if:condition?true:false}`, where the condition evaluation is delayed:

https://github.com/weechat/weechat/blob/3e5d77fdff1f42cd593b5dbd5b585b9cd2830a86/src/core/wee-eval.c#L1883-L1902

Maybe you could do something similar, but this could require to change the format.
For example using: `${hdata_search:search_condition,hdata[list].var1.var2...}`.

Other ideas welcome!

PR #1904, opened 2023-04-08
Fix Lua's `print` from producing a segfault

wfrsk:
fixes https://github.com/weechat/weechat/issues/1905

proof: https://asciinema.org/a/2tGEx0H3N2HRe0gXEjOJdoCHw


flashcode:
Thanks, merged!

Issue #1906, opened 2023-04-08
Calling `debug.debug` on Lua freezes the UI 


advantageous-overtake:
## Bug summary

Window becomes unresponsive after executing `/lua eval debug.debug()`

## Steps to reproduce

1. run `weechat -r "/lua eval debug.debug()"`

## Current behavior

UI becomes unresponsive, resulting in weechat needing to be killed externally.

https://asciinema.org/a/nKWvykOzguKHNDdHiL4C33wIY

## Expected behavior

I don't see why Lua plugins should be able to access the `debug` library, since its always a source of additional issues.

## Suggested solutions

PR https://github.com/weechat/weechat/pull/1907

---

- WeeChat version: 3.8
- OS, distribution and version: Arch Linux
- Terminal: raw tty
- Terminal multiplexer (screen/tmux/…/none): tmux

PR #1907, opened 2023-04-08
Refactor `weechat_lua_load`

advantageous-overtake:
Fixes https://github.com/weechat/weechat/issues/1906

### DIFF
--- a/src/plugins/lua/weechat-lua.c
+++ b/src/plugins/lua/weechat-lua.c
@@ -567,7 +567,6 @@ weechat_lua_load (const char *filename, const char *code)
     luaopen_table     (lua_current_interpreter);
     luaopen_math      (lua_current_interpreter);
     luaopen_os        (lua_current_interpreter);
-    luaopen_debug     (lua_current_interpreter);
 
 #ifdef LUA_VERSION_NUM
 #if   LUA_VERSION_NUM == 501 // Lua 5.1.x
--- a/src/plugins/lua/weechat-lua.c
+++ b/src/plugins/lua/weechat-lua.c
@@ -560,16 +560,34 @@ weechat_lua_load (const char *filename, const char *code)
         return NULL;
     }
 
-#ifdef LUA_VERSION_NUM /* LUA_VERSION_NUM is defined only in lua >= 5.1.0 */
-    luaL_openlibs (lua_current_interpreter);
-#else
-    luaopen_base (lua_current_interpreter);
-    luaopen_string (lua_current_interpreter);
-    luaopen_table (lua_current_interpreter);
-    luaopen_math (lua_current_interpreter);
-    luaopen_io (lua_current_interpreter);
-    luaopen_debug (lua_current_interpreter);
-#endif /* LUA_VERSION_NUM */
+    // Lua 5.0.x libraries
+    luaopen_base      (lua_current_interpreter);
+    luaopen_io        (lua_current_interpreter);
+    luaopen_string    (lua_current_interpreter);
+    luaopen_table     (lua_current_interpreter);
+    luaopen_math      (lua_current_interpreter);
+    luaopen_os        (lua_current_interpreter);
+    luaopen_debug     (lua_current_interpreter);
+
+#ifdef LUA_VERSION_NUM
+#if   LUA_VERSION_NUM == 501 // Lua 5.1.x
+    luaopen_package   (lua_current_interpreter);
+#elif LUA_VERSION_NUM == 502 // Lua 5.2.x
+    luaopen_coroutine (lua_current_interpreter);
+    luaopen_package   (lua_current_interpreter);
+    luaopen_bit32 (lua_current_interpreter);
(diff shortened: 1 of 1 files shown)


flashcode:
Hi,
Tests failed because one message is displayed during lua tests:

    =!= lua: unable to redirect stdout and stderr

Probably caused by your changes.

src/plugins/lua/weechat-lua.c
```diff
@@ -560,16 +560,33 @@ weechat_lua_load (const char *filename, const char *code)
...
+    luaopen_base      (lua_current_interpreter);
+    luaopen_io        (lua_current_interpreter);
+    luaopen_string    (lua_current_interpreter);
+    luaopen_table     (lua_current_interpreter);
+    luaopen_math      (lua_current_interpreter);
+    luaopen_os        (lua_current_interpreter);
+
+#ifdef LUA_VERSION_NUM
+#if   LUA_VERSION_NUM == 501 // Lua 5.1.x
+    luaopen_package   (lua_current_interpreter);
```
flashcode, line 573: Would it be possible to have something cumulative here?
That means: if Lua version ≥ 5.1.x, call `luaopen_package`, so this is done only one time.
Then if Lua version ≥ 5.2.x, call extra functions, etc.

flashcode (changes requested):

src/plugins/lua/weechat-lua.c
wfrsk: It behaves like a switch case, since `LUA_VERSION_NUM` can only be equal to one number at a time.

wfrsk:
Decided to not remove the `debug` library, so it now just sets `debug.debug` to `nil`.

Also, the initialization script used to redirect logs has been replaced by regular Lua API calls.

src/plugins/lua/weechat-lua.c
wfrsk: Reverted, now if `LUA_VERSION_NUM` is NOT defined (compiling against Lua 5.0) it calls each `luaopen_*` function separately, since `luaL_openlibs` does not exist on that Lua version.

If it is defined, it automatically opens all libraries trough `luaL_openlibs`.

wfrsk: > Would it be possible to have something cumulative here?
> That means: if Lua version ≥ 5.1.x, call `luaopen_package`, so this is done only one time.
> Then if Lua version ≥ 5.2.x, call extra functions, etc.

I just noticed i missed your question, sorry!

It would require additional checks, for example:
Lua 5.3 and Lua 5.2 have the bit32 library.
Lua 5.3 added built-in bitwise operators, deprecating the `bit32` library.
Lua 5.4 fully removes the `bit32` library.

Which means something like this would be needed:

```c
(...)
#elif LUA_VERSION_NUM >= 502
#if LUA_VERSION_NUM != 504
    luaopen_bit32 (lua_current_interpreter);
#endif
#elif (...)
```

.gitignore
```diff
@@ -1,6 +1,7 @@
 # ignored files for Git
 
 /build
+/build-tmp-*
```
flashcode, line 4: This seems not related to the changes, could you please remove this?

src/plugins/lua/weechat-lua.c
```diff
@@ -251,19 +251,27 @@ weechat_lua_output_flush ()
...
-weechat_lua_output (lua_State *L)
-{
-    const char *msg, *ptr_msg, *ptr_newline;
-
-    if (lua_gettop (L) < 1)
-        return 0;
-
-    msg = lua_tostring (L, -1);
-    ptr_msg = msg;
+weechat_lua_output (lua_State *L) {
```
flashcode, line 259: Please reformat and use code style described in the doc here: https://weechat.org/files/doc/devel/weechat_dev.en.html#coding_c_style

src/plugins/lua/weechat-lua.c
```diff
@@ -560,32 +558,45 @@ weechat_lua_load (const char *filename, const char *code)
...
-    {
-        weechat_printf (NULL,
-                        weechat_gettext ("%s%s: unable to redirect stdout "
-                                         "and stderr"),
-                        weechat_prefix ("error"), LUA_PLUGIN_NAME);
+    // Remove references to stdout and stderr
+    
+    lua_getglobal(lua_current_interpreter, "io");
+    if (lua_istable(lua_current_interpreter, -1)) {
+        // io.{stdout,stderr} = nil
```
flashcode, line 580: Please remove any commented code.

flashcode (changes requested):

src/plugins/lua/weechat-lua.c
wfrsk: It is not commented code, it is the Lua code "equivalent" of the following API calls, it is considered standard practice when performing API calls.

advantageous-overtake:
Any hints on this being merged?

flashcode:
Hi,
Thanks for the reminder, I'll check again if it's ready for merge.

PR #1908, opened 2023-04-08
core: render newline characters in chat line messages

trygveaa:
If a chat line message contains a newline character (\n) it was previously rendered as J with reverse video. This commit makes it render as an actual newline instead, so messages with multiple lines become supported.

The rendering is fixed in normal mode as well as bare mode both when scrolled to the bottom and when scrolled up (which is different code paths). Focus events has also been updated to support this (except for _chat_line_y which returns -1 for all lines, but the docs says this variable is only for buffers with free content).

Currently, the only way to include a \n in a chat line message is with hdata_update because printf splits on \n and creates multiple separate lines, but hopefully either printf can be changed to not split on \n, or a new command which doesn't split can be added.

### DIFF
--- a/src/gui/curses/gui-curses-chat.c
+++ b/src/gui/curses/gui-curses-chat.c
@@ -488,6 +488,80 @@ gui_chat_display_word_raw (struct t_gui_window *window, struct t_gui_line *line,
 }
 
 /*
+ * Displays the prefix_suffix in the beginning of a line.
+ *
+ * Returns number of chars displayed on screen.
+ */
+
+int
+gui_chat_display_prefix_suffix (struct t_gui_window *window,
+                                struct t_gui_line *line,
+                                const char *word,
+                                int pre_lines_displayed, int *lines_displayed,
+                                int simulate,
+                                int apply_style_inactive,
+                                int nick_offline)
+{
+    char str_space[] = " ";
+    int chars_displayed, length_align;
+
+    chars_displayed = 0;
+
+    /* insert spaces for aligning text under time/nick */
+    length_align = gui_line_get_align (window->buffer, line, 0, 0);
+
+    /* in the beginning of a line */
+    if ((window->win_chat_cursor_x == 0)
+        && (*lines_displayed > pre_lines_displayed)
+        /* FIXME: modify arbitrary value for non aligning messages on time/nick? */
+        && (length_align < (window->win_chat_width - 5)))
+    {
+        /* in the beginning of a word or in the middle of a word with multiline word align */
+        if ((chars_displayed == 0)
+            || CONFIG_BOOLEAN(config_look_align_multiline_words))
+        {
+            window->win_chat_cursor_x += length_align;
+
+            if ((CONFIG_INTEGER(config_look_align_end_of_lines) == CONFIG_LOOK_ALIGN_END_OF_LINES_MESSAGE)
+                && (CONFIG_INTEGER(config_look_prefix_align) != CONFIG_LOOK_PREFIX_ALIGN_NONE)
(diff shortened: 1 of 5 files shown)


trygveaa:
Here's a command I used to test newlines at various places btw:

```
weechat -t -r '/print -escape multi\nline; /print -escape \nnewline at start; /print -escape newline at end\n; /print -escape consecutive\n\nnewlines; /print -escape newline after space \nline; /print -escape space after newline\n line'
```

It requires #1909 to test as well, because it was easier to use `/print` than to write hdata_update commands. But I did verify that it works as expected with hdata_update too (which only requires this PR to test):

```
weechat -t -r '/print a; /print b; /input insert /python eval import weechat as w\; own_lines = w.hdata_pointer(w.hdata_get("buffer"), w.current_buffer(), "own_lines")\; line_pointer = w.hdata_pointer(w.hdata_get("lines"), own_lines, "last_line")\; line_pointer = w.hdata_move(w.hdata_get("line"), line_pointer, -1)\; data = w.hdata_pointer(w.hdata_get("line"), line_pointer, "data")\; w.hdata_update(w.hdata_get("line_data"), data, {"message": "123\\n456"})'
```

pascalpoitras:
suppose you have a multi line message containing three line on a buffer, and you click on the second line, _chat_line_message contains the three line, it would be useful to add a variable to get only the result of the second line

pascalpoitras:
OK finally it will be better to just return the content of the line you selected instead of creating a variable for each line of a multiline message, since multiline message can contains a lot of lines

trygveaa:
I found a small rendering bug with this. If you scroll so a chat line with multiple lines is partially shown, a `|` character is shown at the end of the last message (after `6` in the screenshot):

image

Steps to reproduce:
1. Run `weechat -t -r '/help; /buffer set input_multiline 1; /print -escape 1\n2\n3\n4\n5\n6\n7\n8\n9'`
2. Press alt+pageup to scroll 3 lines up.

I'll try to look into fixing this.

flashcode:
By the way, the MORE indicator displays 0, should it count all lines (multiple lines for a line with some `\n` inside), or buffer "lines", as it is currently?

trygveaa:
I think it makes sense that it only counts buffer "lines". This is also how counting is in the hotlist, it only increases by 1 when a multiline is added. Other people might have different preferences though.

Another thing, `m`, `q` and `Q` in cursor mode includes the whole message with all the lines. This could be useful in some cases, but might also be impractical in other cases. I'm thinking of adding a new key (e.g. `l`) for just the selected line. Agree?

flashcode:
> Another thing, `m`, `q` and `Q` in cursor mode includes the whole message with all the lines. This could be useful in some cases, but might also be impractical in other cases. I'm thinking of adding a new key (e.g. `l`) for just the selected line. Agree?

Yes, good catch and good idea!

trygveaa:
Another thing, `/filter` interprets `\t` as the separator between prefix and message, but doesn't interpret `\n`. Matching newlines in a filter works if you write a literal newline character though. So like this:

```
/eval /filter addreplace test * * a${\n}b
```

or

```
/filter addreplace test * * a
b
```

Both only works if run from a buffer with input_multiline set.

I think we should fix so you can add the filter like this instead (which would work from any buffer, not just the input_multiline ones):

```
/filter addreplace test * * a\nb
```

Side note: I can't actually get `\n` in a filter regex to match anything currently. I have to use `\\n` to match a message containing a literal `\n` (which makes sense, just a bit weird then that `\n` doesn't work).

flashcode:
Thanks, merged!

PR #1909, opened 2023-04-08
core: don't split on newline characters in printf_date_tags

trygveaa:
With support for rendering newline characters as new lines, we don't need to split the message on newline characters anymore in printf_date_tags. This allows you to print a line with multiple lines.

Depends on #1908.

### DIFF
--- a/src/gui/gui-chat.c
+++ b/src/gui/gui-chat.c
@@ -844,7 +844,6 @@ gui_chat_printf_date_tags (struct t_gui_buffer *buffer, time_t date,
                            const char *tags, const char *message, ...)
 {
     time_t date_printed;
-    char *pos, *pos_end;
 
     if (!message)
         return;
@@ -867,25 +866,14 @@ gui_chat_printf_date_tags (struct t_gui_buffer *buffer, time_t date,
     if (date <= 0)
         date = date_printed;
 
-    pos = vbuffer;
-    while (pos)
+    if (gui_init_ok)
     {
-        /* display until next end of line */
-        pos_end = strchr (pos, '\n');
-        if (pos_end)
-            pos_end[0] = '\0';
-
-        if (gui_init_ok)
-        {
-            gui_chat_printf_date_tags_internal (buffer, date, date_printed,
-                                                tags, pos);
-        }
-        else
-        {
-            gui_chat_add_line_waiting_buffer (pos);
-        }
-
-        pos = (pos_end && pos_end[1]) ? pos_end + 1 : NULL;
+        gui_chat_printf_date_tags_internal (buffer, date, date_printed,
+                                            tags, vbuffer);
+    }
+    else
+    {
+        gui_chat_add_line_waiting_buffer (vbuffer);
     }
 
     free (vbuffer);
--- a/src/gui/gui-chat.c
+++ b/src/gui/gui-chat.c
@@ -844,6 +844,8 @@ gui_chat_printf_date_tags (struct t_gui_buffer *buffer, time_t date,
                            const char *tags, const char *message, ...)
 {
     time_t date_printed;
+    char *pos, *pos_end;
+    int one_line = 0;
 
     if (!message)
         return;
@@ -866,14 +868,37 @@ gui_chat_printf_date_tags (struct t_gui_buffer *buffer, time_t date,
     if (date <= 0)
         date = date_printed;
 
-    if (gui_init_ok)
+    pos = vbuffer;
+    while (pos)
     {
-        gui_chat_printf_date_tags_internal (buffer, date, date_printed,
-                                            tags, vbuffer);
-    }
-    else
-    {
-        gui_chat_add_line_waiting_buffer (vbuffer);
+        if (!buffer || !buffer->input_multiline)
+        {
+            /* display until next end of line */
+            pos_end = strchr (pos, '\n');
+            if (pos_end)
+                pos_end[0] = '\0';
+        }
+        else
+        {
+            one_line = 1;
+        }
+
+        if (gui_init_ok)
+        {
+            gui_chat_printf_date_tags_internal (buffer, date, date_printed,
+                                                tags, pos);
+        }
+        else
+        {
+            gui_chat_add_line_waiting_buffer (pos);
+        }
+
+        if (one_line)
+        {
+            break;
+        }
+
+        pos = (pos_end && pos_end[1]) ? pos_end + 1 : NULL;
     }
 
     free (vbuffer);


trygveaa:
Some considerations have to be taken regarding aligning of lines. With this the WeeChat ascii logo is rendered as a single line, which with #1908 as it currently is means that it won't be aligned if `weechat.look.align_end_of_lines` is changed to anything other than `message`, and it won't be aligned in bare mode and logs. Maybe we could add alignment to bare mode and logs, but with `weechat.look.align_end_of_lines` set to anything other than `message` it won't be aligned as that's what the option is. The logo could also be changed to be printed as multiple lines if this is the only alignment that matters.

pascalpoitras:
so if a trigger call a command, for example /print, on tg_message or tg_message_nocolor and this message contain a newline (possible on slack, I guess matrix too?) someone could maliciously do something like "hello(newline)/quit" to make the user quit or something nastier like "hello(newline)/eval ${sec.data.yourpass}" ?

so any trigger using such previous valid thing would have to be rewritten to avoid injection ?

a newline make /print sent and then execute what follow the newline as another command

`/trigger add test print "core.test" "" "===.*===hello\n/print -core malicious code goes here===my_text" "/print -buffer core.test ${my_text}" `

then

`/print -newbuffer test hello`

flashcode:
@pascalpoitras: then we should also change the behavior of `command` API function (called by trigger to execute the command): there should be no split on newline, so the string is always evaluated as a single command.

This could be done as a separate commit though, not directly related to these changes.

trygveaa:
Oh, good catch. One thing to note is that this is only a problem on buffers which haven't set the `input_multiline` buffer property. On buffers with that set (e.g. slack buffers have this set), the whole message including newlines is sent to the command, so nothing will be executed.

I don't think your trigger example makes sense to demonstrate the problem since you have the malicious code in the trigger itself. If you run that on 3.8 you'll see the same behavior as in this PR. However, the problem you describe is real, and can be demonstrated with these commands:

```
/trigger addreplace test print "" "" "" "/print \tmsg: ${tg_message}"
/print -escape 1\n/eval -n ${irc.server_default.ssl_priorities}
```

Which will print this with 3.8:

```
23:28 1
23:28 msg: 1
23:28 /eval -n ${irc.server_default.ssl_priorities}
23:28 msg: /eval -n ${irc.server_default.ssl_priorities}
```

And this with this PR with `input_multiline` set:

```
23:31 1
/eval -n ${irc.server_default.ssl_priorities}
23:31 msg: 1
/eval -n ${irc.server_default.ssl_priorities}
```

But this with this PR and `input_multiline` not set:

```
23:31 1
/eval -n ${irc.server_default.ssl_priorities}
23:31 msg: 1
23:31 >> ${irc.server_default.ssl_priorities}
23:31 == [NORMAL:-VERS-SSL3.0]
```

> so any trigger using such previous valid thing would have to be rewritten to avoid injection?

No, we should fix the PR so injection isn't possible.

So there's two possible fixes. Either escape newline characters when they're sent to the trigger command, or make trigger commands always behave as if `input_multiline` is set (the latter is what @flashcode suggests). I agree that it makes sense to not split the lines for the trigger command.

trygveaa:
> This could be done as a separate commit though, not directly related to these changes.

Since this introduces a potential attack vector, we should probably fix it as part of this change, not separately (The related #1908 shouldn't be a problem though, just this PR).

> So there's two possible fixes. Either escape newline characters when they're sent to the trigger command, or make trigger commands always behave as if input_multiline is set (the latter is what @flashcode suggests). I agree that it makes sense to not split the lines for the trigger command.

Another alternative is to keep splitting lines in printf for buffers without `input_multiline` set, and only drop the split for buffers with that set. A bit weird to tie this into something named `input_...`, but could rename it.

pascalpoitras:
thanks a lot trygveaa for all this information, it is very technical to me so I'll let you and flashcode figure how to fix this the better way, thanks again

flashcode:
I think your last proposition is the best:

- if `input_multiline` is not set, behave like before: split on `\n` in printf and command
- if `input_multiline` is set, don't split on `\n`.

That way no attack should be possible.

I'm not sure it's worth renaming the `input_multiline`, it's somewhat internal and should be enabled by scripts only.

And I agree this should be fixed at same time (I mentioned separate commit, but my idea was to push everything together).
Anyway if this solution is chosen, this PR must be updated.

pascalpoitras:
as @trygveaa told me on IRC, input_multiline property is already set for slack but for matrix is it ? if not, we will have problem on 4.0.0 with matrix if we not set this property explicitly? (and even if matrix did it like slack there can be another protocol that doesn't)

so I think input_multiline property should be set by default for buffers (otherwise I'll have to reach everyone using my conf to set this property)

flashcode:
So another proposition: remove completely `input_multiline` and enable it always.

Then as the IRC protocol doesn't support multiline in messages (yet?), the irc plugin could split on `\n` to send multiple messages when the input has `\n` inside.
And commands would never be split on `\n`.

flashcode:
This way multiline is default, and IRC protocol is the exception.

What do you think @trygveaa, do you see some side effects on doing this?

flashcode:
One idea to address this problem:

- the `weechat_command` API function never splits on `\n` → no attack via triggers
- on manual user input (ie when command `/input return` is executed, which is bound to `Enter` key by default):
  - if the first line is a command, auto-split on `\n` to send multiple commands
  - if the first line is NOT a command, don't split and send the whole input as a single message

That way, a series of commands copy/pasted into the input would execute multiple commands.

I think it's uncommon for a user to send a command that contain real newline chars.

What do you think @trygveaa?

pascalpoitras:
@flashcode just a question

if the trigger is a modifier input_text_content, input_text_display,  input_text_display_with_cursor or  input_text_for_buffer can it cause problem? I know you just said no attack via triggers but still they manipulate the input

trygveaa:
I think that sounds like a good solution, except that we need to keep the possibility of running commands that has multiple lines in some way, as wee-slack is currently using this. Commands with multiple lines is useful because sometimes you need to send a message using a command rather than just typing it (think `/msg` in IRC, in wee-slack it's the `/reply` command), and messages can have multiple lines.

Currently this works in wee-slack since its buffers has `input_multiline` set (which means that you currently can't paste multiple newline separated commands in wee-slack buffers).

We could do it by keeping the `input_multiline` property, which would then only control the behavior for commands. Alternatively, the command created by the script could specify that it supports/needs multiple lines, so if WeeChat sees such a command, it should not split.

flashcode:
Ah I see, so the auto-split is not ideal as well.

Having the split behavior set in the command is maybe not even good: for example the `/print` command could allow newlines, but what if you want to send multiple `/print` with one line for each? WeeChat would then not split as requested by the command itself, and would not execute the subsequent print commands.

trygveaa:
Specifically for `/print`, we could add an option to specify that each line should be a separate chat line. It would mean that if you have any commands on subsequent lines, they would be printed rather than executed though.

We could also add a command which splits on `\n` before executing, e.g.:

```
/command -split * /print 1
/print 2
```

Would run `/print 1` and `/print 2`. Maybe this can solve such edge cases? We should probably not require using this command for multiple lines of e.g. `/set` commands though, as that would be hard for users to discover, especially given all the examples already on web pages with commands to paste into WeeChat.

flashcode:
> @trygveaa: I think your last proposition is the best:
> 
> * if `input_multiline` is not set, behave like before: split on `\n` in printf and command
> * if `input_multiline` is set, don't split on `\n`.
> 
> That way no attack should be possible.
> 
> I'm not sure it's worth renaming the `input_multiline`, it's somewhat internal and should be enabled by scripts only.
> 
> And I agree this should be fixed at same time (I mentioned separate commit, but my idea was to push everything together). Anyway if this solution is chosen, this PR must be updated.

So maybe this is the best solution, isn't it?

trygveaa:
> So maybe this is the best solution, isn't it?

It's at least the closest one to the current behavior, and thus perhaps the safest. Though if (when?) IRC gets multiline support, the core buffers would be the only with `input_multiline` not set, and perhaps it would be confusing for users that commands work differently in different buffers.

New suggestion: What if we never split commands/input like you suggested above (as if `input_multiline` is always set), and to support multiple commands we detect it either when pasting or when running the command (triggered when all lines start with `/`), and show a prompt to the user asking if they want to execute multiple commands or not?

trygveaa:
> > So maybe this is the best solution, isn't it?
> 
> It's at least the closest one to the current behavior, and thus perhaps the safest. Though if (when?) IRC gets multiline support, the core buffers would be the only with `input_multiline` not set, and perhaps it would be confusing for users that commands work differently in different buffers.

Also, it would be a problem if `input_multiline` is set on a buffer, then multiline chat lines are printed and then `input_multiline` is turned off. I don't have a use case for this though (wee-slack doesn't turn if off), so we could disallow turning it off if we want to be sure.

trygveaa:
> if the trigger is a modifier input_text_content, input_text_display, input_text_display_with_cursor or input_text_for_buffer can it cause problem? I know you just said no attack via triggers but still they manipulate the input

If you have an `input_text_content` modifier, it would replace the input contents so you would see what the input is before potentially running it, and thus shouldn't be a problem. Presumably, you wouldn't have a trigger which runs `/input return` automatically?

`input_text_display` and `input_text_display_with_cursor` only changes how the text is displayed in the input, so it shouldn't be a problem. Unless it removes the newlines so you think you're running one command, but actually ends up running multiple. But I think that would only happen if the trigger is actively malicious or buggy, so probably not a problem. Either way, not relevant here since if this is a problem, the problem exists in the released versions as well.

`input_text_for_buffer` could be a problem though, since it changes the input after the user runs it and before it's interpreted as commands. If you have a trigger which inserts a message with newlines here, and we split the input and runs it as separate commands it's problematic. If this is a problem or not depends on the solution we go for though. It shouldn't be a problem with the suggestion two comments up to prompt the user if they want to execute multiple commands.

flashcode:
THinking about the proposed solution: that also means on buffers with `input_multiline` enabled, we would not be able to paste multiple commands at once and execute them separately.
For now this would still be possible on IRC buffers, but maybe not in future when multiline will be enabled there.

Also, why turning off `input_multiline` is an issue?

trygveaa:
> Thinking about the proposed solution: that also means on buffers with `input_multiline` enabled, we would not be able to paste multiple commands at once and execute them separately.

Yes, this is already the case actually. It could be possible to support with a command though, [like I suggested here](https://github.com/weechat/weechat/pull/1909#issuecomment-1503995181).

> Also, why turning off `input_multiline` is an issue?

If `input_multiline` is on and e.g. `something\n/command` is printed, and you turn off `input_multiline` and then run a trigger on that line, the command will be executed. I guess there aren't many triggers operating on lines that already has been printed, but it's at least the focus trigger.

This also happens if a line is updated to include a newline character with hdata_update. Here is a demonstration of this which shows that happening in 3.8 as well:

```
weechat -t -r '/mouse enable; /trigger addreplace print_focus focus chat "" "" "/print \${_chat_line_message}"; /print a; /python eval import weechat as w\; own_lines = w.hdata_pointer(w.hdata_get("buffer"), w.current_buffer(), "own_lines")\; line_pointer = w.hdata_pointer(w.hdata_get("lines"), own_lines, "last_line")\; data = w.hdata_pointer(w.hdata_get("line"), line_pointer, "data")\; w.hdata_update(w.hdata_get("line_data"), data, {"message": "click here\n/eval -n \${buffer.name}"})'
```

Click on `click here` and see the command be executed.

flashcode:
Another possible solution to force split would be to add a new key (I have no idea which one would be fine, as meta-return is already used) that would split and send each line separately, possibly executing multiple commands.

trygveaa:
Okay, so the current standing proposal is that we'll do the change in this PR only for buffers with `input_multiline` set instead of for all buffers.

So for buffers without `input_multiline` set, nothing changes.

For buffers with `input_multiline` set, there won't be any split on newline on any print (with `/print` or from an API function) or command execution. This means that you can print a chat line containing multiple lines only from such buffers. It won't be possible to execute multiple commands by pasting them to the input bar, you have to run them one by one, but this is already the case for such buffers. We can later add a separate key to run multiple commands for these buffers like @flashcode suggested.

This addresses the mentioned issues with triggers since either there is no newline characters (for input_multiline=0) or there will be no split (for input_multiline=1) so only the first command is executed. This goes for all triggers, print, focus, modifier etc.

There are two small issues remaining with this suggestion:

1. If hdata_update is used to add a newline to a chat line in a buffer without input_multiline set, the issues mentioned above remains.
2. If a buffer with input_multiline set has chat lines with newlines in them, and then input_multiline is turned off, the issues mentioned above remains.

I don't think these issues are likely to be a problem, because there wouldn't be any reason for script authors do to these things. But if we want to address them we could prevent using a newline character in hdata_update if input_multiline isn't set, and we could prevent turning off input_multiline after it's been turned on (or prevent turning it off if any chat lines has newline characters).

pascalpoitras:
I still see a potential command injection

suppose I have a channel monitor like this                                                                                                                                    
                                                                                                                                                                              
`/trigger add chanmon print "*;irc_privmsg" "" "" "/print -newbuffer chanmon \t${tg_message}"    `                                                                              
                                                                                                                                                                              
then suppose I am on a buffer with input_multiline set to 1 (for example on slack)                                                                                            
                                                                                                                                                                              
someone wrote something so the trigger is executed                                                                                                                            
the message contains some newlines in it so tg_message will contains newlines                                                                                                 
then the trigger execute the /print command but the chanmon buffer input_multiline is set to 0                                                                                
what will happens ? command injection.                                                                                                                                        
                                                                                                                                                                              
For example if the user on slack type "hello\n/eval ${sec.data.pass}"                                                                                                         
the \n is keep in tg_message since the slack buffer input_multiline is set to 1                                                                                               
but then we print on another buffer (chanmon) with input_multiline set to 0                                                                                                   
                                                                                                                                                                              
So you might say, just set input_multiline to 1 on chanmon too, sure but every user that is currently (as we speak) using such a trigger will have to set the property to 1...

The problem here... is to contact everyone using such a trigger to tell him to set the input_multiline buffer property to 1 and if you can't then someone could inject command

flashcode:
This is not completely true: trigger executes the command `/print -newbuffer chanmon \t${tg_message}` on the buffer where the `print` trigger is running, so on the slack buffer in your case.

That means if this buffer has `input_multiline` set to 1, the message is *NOT* split.
Then the command is executed and the `/print` command will display it on another buffer, no problem if `input_multiline` is not set there: this command just prints a message, the text is *NOT* sent to a buffer (ie it can not be split).

So I think there's no possible code injection there.

Though this has to be checked against new changes to be sure.

trygveaa:
Yes, @flashcode is right. I tested it, and it behaves like he described, so no command from the message is run.

trygveaa:
I pushed a commit to only avoid splitting in printf in buffers with input_multiline set: https://github.com/weechat/weechat/pull/1909/commits/60f1b4b17d52c2b4b0ac1199292b527f8dd78213

This is the commit I tested in the previous comment.

pascalpoitras:
ah yes you're right the /print is executed on the buffer where the trigger was executed

trygveaa:
I saw another thing. The response from `/exec` is still printed as multiple lines even when the buffer has input_multiline set. E.g.:

```
/exec -norc printf "hello\nworld"
```

Will add two new chat lines, instead of just one. I think we should change it to only add one chat line with the whole response.

flashcode:
I'm not sure why output of /exec should be displayed as a single line.
What if the output is big, say 100 or 1000 lines, display a single line in WeeChat?

flashcode:
Thanks, merged!

trygveaa:
> I'm not sure why output of `/exec` should be displayed as a single line.

I noticed this when testing [@pascalpoitras' translate trigger](https://gist.github.com/pascalpoitras/8406501#translate-with-translate-shell) and saw that if you run it on a multiline message it will print the translated result as multiple lines, rather than the single line the original message was.

There can be other reasons as well. E.g. if you have a desktop notification on the result you will get one notification for each line instead of a single notification (maybe far fetched to have a notification on the result of `/exec` though).

I also though that since the lines of the response belong together it would make sense to have it as a single line.

There probably can be use cases for having it as separate lines as well though, depending on what you run with `/exec`.

> What if the output is big, say 100 or 1000 lines, display a single line in WeeChat?

Is that a problem?

flashcode:
Maybe not a problem, if very long lines are properly handled and displayed.

trygveaa:
I can see reasons for keeping it as multiple chat lines though. E.g. if you want to filter parts of the output. So one option is to add an option to `/exec` to choose if the response is split or not.

pascalpoitras:
similar to hdata_update issue above, if someone use a trigger or script with hook_line and that script or trigger add newline to the message, then my print trigger which '/print ...' will be vulnerable on an input_multiline buffer set to 0

Also, if you have a hook_command trigger, it might behave differently if you run that command on a input_multiline buffer or not (if there's \n pass to the command)

For these reason it would be nice to add a new option for trigger which depending on the value:

look at the buffer's input_multiline property to know if it should cut or not (current behavior)
always cut (like if the buffer input_multiline property was set to 0 even if it's not)
never cut (like if the buffer input_multiline property was set to 1 even if it's not)

by default it would behave as it behave currently (depending on the buffer input_multiline property)

flashcode:
Thanks for reporting the issue, I fixed the hook_line to keep only first line if the target buffer has `input_multiline` set to 0 (like it's already done for hdata update of a line).

I'm not fan of adding a way to force/prevent cut on newlines in trigger, I think it's better that the trigger author doesn't have to do this and fix any function outside trigger, that could add newlines where it's not permitted.

pascalpoitras:
what about the hook_modifier, this modifier change the tg_string variable

pascalpoitras:
For the hook_line, the problem is not just with the message variable. The problem is with all the variables that are updated

buffer, buffer_name, y, date, date_printed, str_time, tags, notify_level, highlight, prefix, message

why? because another script or trigger author might create a trigger which add a newline in these variables

so suppose I have a perfectly secure trigger, I gave it to someone who is using a script or a trigger which is not mine and put a \n in these var, then my trigger is sensible to code injection

this is not a problem for the variable that are not updated because you must put the \n in the same trigger and trigger author should know that this is not a good idea. 

With updated var you can't predict what other author will do that will affect your trigger

what do you think @trygveaa @flashcode about protecting the other updated vars of hook_line

flashcode:
Hi @pascalpoitras, @trygveaa,

I pushed this branch for tests: https://github.com/weechat/weechat/tree/command-no-split-newline-by-default

The changes:

- in function `weechat.command()`, there is no split on `\n` any more if `input_multiline` = 0 (if multiple lines are received, only the first line is used, the rest is ignored) ;
- in function `weechat.command_options()`, if the hashtable variable `split_newline` is set to `1` (string "1"), the split is done, and then multiple commands can be executed at once ;
- no changes when `input_multiline` = 1: the whole string including `\n` is used.

@trygveaa: do you see potential problems with these changes? (other than possibly breaking scripts that use `weechat.command()` to send multiple commands at once, separated by `\n`)?

pascalpoitras:
@trygveaa a hook_line update the following variables: buffer, buffer_name, y, date, date_printed, str_time, tags, notify_level, highlight, prefix, message
so someone can add a newline in them after the split on newline is done (on an input_multiline buffer) and then weechat.command will split on newline

trygveaa:
Right, since the command execution in hook_print (and other hook/triggers) uses weechat.command. Thanks for the explanation.

Then that fix sounds great.

PR #1914, opened 2023-04-24
Add _chat_focused_line variable to get the focused line

### LINKED ISSUE #1913: Add variable with current line of message when input_multiline is set
## Feature description

As requested in this [comment](https://github.com/weechat/weechat/pull/1908#issuecomment-1515321949), a new variable `_chat_line_line` should be added in focus hashtable, containing the content of current line in the message.

The current line can be different from the message when `input_multiline` is set to 1, and some `\n` are displayed in the message.

In addition, the key `l` (lower L) should be added in cursor mode to quote the current line of the message.

trygveaa:
This contains the single line that you focused on, like _chat_word
contains the word that you focused on. This will be equal to
_chat_line_message if the chat line only contains a single line, but if
the chat line has multiple lines, _chat_line_message will contain all of
them, but _chat_focused_line will only contain the single line that was
focused.

Also adds cursor key l to quote the focused line.

Fixes #1913


trygveaa:
@flashcode: I see you chose the name `_chat_line_line` in #1913. It seemed a bit weird to me to have line twice (it's a bit unfortunate that line now means two different things), and I had already chosen `_chat_focused_line`, so I kept that. But I'm not sure what's the best name.

PR #1915, opened 2023-04-25
ruby: Restore default SIGCHLD signal handler on unload

### LINKED ISSUE #1889: Segfault when running /quit with an ongoing exec
## Bug summary

If you run `/quit` while WeeChat is running a process with `/exec` or `hook_process`, WeeChat will segfault.

## Steps to reproduce

Easiest way is just running: `weechat -t -r '/exec sleep 5; /quit'`

It can also be reproduced by having a script that uses `hook_process` and running `/quit` while a process is ongoing. This can be any process, a command like above, a URL or a function.

## Additional information

Backtrace from gdb:

```
Program terminated with signal SIGSEGV, Segmentation fault.
#0  0x00007f05051d0870 in ?? ()
[Current thread is 1 (Thread 0x7f0506de0940 (LWP 557808))]
(gdb) bt full
#0  0x00007f05051d0870 in ?? ()
No symbol table info available.
#1  <signal handler called>
No locals.
#2  0x00007f05051d1060 in ?? ()
No symbol table info available.
#3  <signal handler called>
No locals.
#4  0x00007f0507b05747 in __GI___wait4 (pid=557816, stat_loc=0x0, options=0, usage=0x0)
    at ../sysdeps/unix/sysv/linux/wait4.c:30
        sc_ret = 557816
        sc_cancel_oldtype = 0
        sc_ret = <optimized out>
#5  0x000055a35a332dc7 in hook_process_free_data (hook=0x55a35b9298f0)
    at /home/trygve/dev/weechat/src/core/hook/wee-hook-process.c:863
No locals.
#6  0x000055a35a31045c in unhook (hook=0x55a35b9298f0)
    at /home/trygve/dev/weechat/src/core/wee-hook.c:544
No locals.
#7  0x00007f0505d80adb in exec_free (exec_cmd=0x55a35bec5030)
    at /home/trygve/dev/weechat/src/plugins/exec/exec.c:638
        i = 21923
#8  0x00007f0505d80bfc in exec_free_all () at /home/trygve/dev/weechat/src/plugins/exec/exec.c:669
No locals.
#9  0x00007f0505d8128c in weechat_plugin_end (plugin=0x55a35b984720)
    at /home/trygve/dev/weechat/src/plugins/exec/exec.c:788
No locals.
#10 0x000055a35a38ce4f in plugin_unload (plugin=0x55a35b984720)
    at /home/trygve/dev/weechat/src/plugins/plugin.c:1245
        end_func = 0x7f0505d8126c <weechat_plugin_end>
        name = 0x55a35bd82270 "exec"
#11 0x000055a35a38cfa8 in plugin_unload_all () at /home/trygve/dev/weechat/src/plugins/plugin.c:1297
        plugins_loaded = 1
#12 0x000055a35a38d2f1 in plugin_end () at /home/trygve/dev/weechat/src/plugins/plugin.c:1417
No locals.
#13 0x000055a35a2d313e in weechat_end (gui_end_cb=0x55a35a384277 <gui_main_end>)
    at /home/trygve/dev/weechat/src/core/weechat.c:675
No locals.
#14 0x000055a35a2d2083 in main (argc=4, argv=0x7fffe610af38)
    at /home/trygve/dev/weechat/src/gui/curses/normal/main.c:44
No locals.
```

---

- WeeChat version: Backtrace is from master (a5aa3304a), also happens in latest stable (v3.8)
- OS, distribution and version: Arch Linux
- Terminal: kitty
- Terminal multiplexer (screen/tmux/…/none): none

talisein:
fixes #1889

I can't explain why some people can't reproduce #1889. Perhaps the order their plugins are unloaded is different, or there is another plugin that is installing yet another SIGCHLD signal handler.

I've examined ruby's signal.c from 3.2.2, 2.7.4, and 3.1.2; the issue should be happening in across all versions.

In any case, this change sets the SIGCHLD handler to the default handler, which should be a pretty safe thing to do in all cases. The manual for signal(2) says `The only portable use of signal() is to set a signal's disposition to SIG_DFL or SIG_IGN.` so I guess this should be fine even on BSD; but I've only tested on linux-x86_64

### DIFF
--- a/src/plugins/ruby/weechat-ruby.c
+++ b/src/plugins/ruby/weechat-ruby.c
@@ -30,6 +30,7 @@
 #include <ruby/version.h>
 #endif
 
+#include <signal.h>
 #include <stdarg.h>
 #include <sys/types.h>
 #include <sys/stat.h>
@@ -1440,6 +1441,7 @@ weechat_plugin_end (struct t_weechat_plugin *plugin)
     ruby_quiet = 0;
 
     ruby_cleanup (0);
+    signal (SIGCHLD, SIG_DFL);
 
     /* free some data */
     if (ruby_action_install_list)


flashcode:
Thanks, merged.

PR #1943, opened 2023-05-28
core: fix chat colors at certain positions not being applied

### LINKED ISSUE #1928: IRCv3 draft/multiline starting another line than the first with color doesn't work
## Bug summary

If any lines other than the first start with a color code, it doesn't work

## Steps to reproduce

1. use ergo testnet
2. type the following in the input bar (in a channel)
3. <ctrl-c c 4>hello
4. alt+enter
5. <ctrl-c c 5>test
6. enter

## Current behavior

the second line is in color 4 too but it should be 5. It works if the color is not the first thing in the line tho

---

- WeeChat version: 4.0.0-dev
- OS, distribution and version: Debian 11
- Terminal: kitty
- Terminal multiplexer (screen/tmux/…/none): tmux

trygveaa:
The new rendering of multiline lines had some issues with colors at certain positions not being applied. The color would not be applied if the color code was at either of these positions:

  - At the start of a line after a newline character
  - At the end of a line after a space and before a newline character
  - At a line by itself before a newline character

The way I had done it by considering newline characters as a word in gui_chat_get_word_info with a variable specifying that it's newline characters became messy and didn't really make sense, so rather than doing this, I changed gui_chat_get_word_info to stop before the first newline character. That way, we can just check if we are at a newline character at the start of the loop, and don't need any more special handling.

Fixes #1928

### DIFF
--- a/src/gui/curses/gui-curses-chat.c
+++ b/src/gui/curses/gui-curses-chat.c
@@ -1544,7 +1544,7 @@ gui_chat_display_line (struct t_gui_window *window, struct t_gui_line *line,
 
             /* if message ends with spaces, display them */
             if ((word_length <= 0) && (word_length_with_spaces > 0)
-                && !ptr_data[word_end_offset + 1])
+                && !ptr_data[word_end_offset])
             {
                 word_length = word_length_with_spaces;
             }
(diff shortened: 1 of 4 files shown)


flashcode:
Thanks!

PR #1947, opened 2023-05-30
core: fix weechat.look.align_multiline_words not working

trygveaa:
This option was broken with commit 031bd45e3.


flashcode:
Thanks!

PR #1964, opened 2023-06-24
doc: fix URL to WeeChat documentation

adsr:


flashcode:
Thanks for the PR, instead I fixed weechat.org site to allow pointing to the project doc without mentioning "stable" or "devel" ("stable" being the default value): https://github.com/weechat/weechat.org/commit/83b56c0eb0d3d1257e8f283dff13908e03ce2691

PR #1968, opened 2023-06-26
core: check for newline character in string_is_whitespace_char

trygveaa:
This fixes a bug where if you had multiple lines in the input and pressed ctrl-w when the cursor was after the first word of any line but the first, it would delete both the word before the cursor and the last word on the preceding line.

### DIFF
--- a/src/core/wee-string.c
+++ b/src/core/wee-string.c
@@ -1312,7 +1312,7 @@ string_convert_escaped_chars (const char *string)
 }
 
 /*
- * Checks if first char of string is a whitespace (space or tab).
+ * Checks if first char of string is a whitespace (space, tab, newline or carriage return).
  *
  * Returns:
  *   1: first char is whitespace
@@ -1322,7 +1322,11 @@ string_convert_escaped_chars (const char *string)
 int
 string_is_whitespace_char (const char *string)
 {
-    return (string && ((string[0] == ' ') || string[0] == '\t')) ? 1 : 0;
+    return (string && (
+            (string[0] == ' ')
+            || (string[0] == '\t')
+            || (string[0] == '\n')
+            || (string[0] == '\r'))) ? 1 : 0;
 }
 
 /*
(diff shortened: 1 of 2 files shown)


src/core/wee-string.c
```diff
@@ -1322,7 +1322,10 @@ string_convert_escaped_chars (const char *string)
 int
 string_is_whitespace_char (const char *string)
 {
-    return (string && ((string[0] == ' ') || string[0] == '\t')) ? 1 : 0;
+    return (string && (
+            (string[0] == ' ')
+            || (string[0] == '\t')
+            || (string[0] == '\n'))) ? 1 : 0;
```
flashcode, line 1328: What about adding `\r` as well here?

trygveaa: Yeah, that makes sense. I took a look at https://en.wikipedia.org/wiki/Whitespace_character and there's actually a lot of characters we should probably check for here. But maybe not worth to check for all of those?

E.g. non-breaking space (U+00A0) may be worth checking for though? Can I assume that the string is in UTF-8 if adding a check for that?

trygveaa: Can't we use the `isspace` function here instead of checking for each character individually?

That one doesn't support unicode though. There is `iswspace`, but not sure exactly how to use it in this context. I tried `return string && iswspace (utf8_char_int (string));`, but it didn't detect U+00A0 as a white space character.

flashcode: As you're targeting branch 4.0 in this PR, I would suggest to just add the strict minimum, ie we can just add `\r`.

On master branch, we can think about a different solution that takes care about other chars.

trygveaa: > As you're targeting branch 4.0 in this PR, I would suggest to just add the strict minimum, ie we can just add `\r`.
> 
> On master branch, we can think about a different solution that takes care about other chars.

Sure, makes sense. I changed it to include `\r` now.

> By the way bash doesn't recognize U+00A0 as a space.

Yeah, but e.g. vim does. Don't think we should necessarily limit us to what bash does.

trygveaa: But people probably rarely write non breaking space and other white space characters than the ones we check here, so might not be worth spending time into.

PR #1979, opened 2023-07-09
Fix bar_item_col and bar_item_line for wrapped/new lines with horizontal filling

trygveaa:
bar_item_col:

When the bar item filling is horizontal and the text of an item is
either wrapped or contains a \r so it's multiple lines and you focus on
a line that isn't the first, the bar_item_col value was incorrect. It
only accounted for focuses on the first line and set the col based on
where the item starts on that line, but for subsequent lines they start
at the beginning of the bar which is often a different column.

bar_item_line:

When the bar item filling is horizontal and the text of an item is
either wrapped or contains a \r so it's multiple lines and you focus on
a line that isn't the first, the bar_item_line value did not represent
the line you focused on in the bar item.

Instead, it represented which line the character you focused on
originally was in the string used to create the bar item. This is
different because \n in this string is converted to a space in a bar
item, so only \r actually creates a new line.

In my opinion, it's much more useful to know which line you focused on
in the bar item, and this is what the variable name sounds like it
represents. So the variable is changed to that.

Note that the coords->line variable is now unused. If it's useful, we
could assign it to a new variable in the focus hashtable, or if not we
should probably remove it.

### DIFF
--- a/src/gui/gui-bar-window.c
+++ b/src/gui/gui-bar-window.c
@@ -224,7 +224,8 @@ gui_bar_window_search_by_xy (struct t_gui_window *window, int x, int y,
                     {
                         *bar_item = (*bar_window)->bar->items_name[item][subitem];
                         *bar_item_line = (*bar_window)->coords[i]->line;
-                        *bar_item_col = x - (*bar_window)->coords[i]->x;
+                        *bar_item_col = x - (y == (*bar_window)->coords[i]->y ?
+                            (*bar_window)->coords[i]->x : (*bar_window)->x);
                     }
                     break;
                 }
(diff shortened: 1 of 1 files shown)


trygveaa:
Maybe I should have opened this against the 4.0 branch?

trygveaa:
Turns out I misunderstood these variables, so we shouldn't change them like this. They should represent the position of the character you clicked in the original string, not the position you clicked on the bar item. After @flashcode pointed out that buflist uses `_bar_item_line` I realized how this is used (when you set buflist to horizontal, the lines are merged, but you need to know which line you clicked on to know which buffer you clicked on).

As far as I can see, the value of `_bar_item_col` is incorrect for wrapped lines or after \r for items that doesn't start at the beginning of the item though. But the change here isn't the correct fix.

For my use case, which is implementing support for clicking on the input bar to move the cursor where you clicked, we should rather introduce the variables `_bar_item_x` and `_bar_item_y` with the values in this PR. However, it turned out that supporting clicking on the input bar was more work than I anticipated when it comes to supporting wrapped and scrolled lines, so not sure if/when I'll continue with it.

PR #1988, opened 2023-07-27
fset: fix fset.color.help_description not applying

nathanielevan:
**Description**

**[why]**
The fset.color.help_description setting previously didn't apply to its corresponding UI element, in this case the description of configuration variable in the help bar in fset.

**[how]**
Change code that configures the foreground colour of this UI element to apply the aforementioned setting.


nathanielevan:
Quick demonstration of how this looks like:
| old | new |
|-|-|
|scshot-310723-212620|scshot-310723-232820|

flashcode:
Tanks!

PR #2021, opened 2023-09-24
Fix few endianness issues

luk1337:
These issues were spotted when running tests on s390x system.

### DIFF
--- a/src/core/wee-crypto.c
+++ b/src/core/wee-crypto.c
@@ -524,6 +524,10 @@ weecrypto_totp_generate_internal (const char *secret, int length_secret,
     int rc, offset, length;
     unsigned long bin_code;
 
+#if __BYTE_ORDER == __BIG_ENDIAN
+    /* Big endian does not need to swap bytes here! */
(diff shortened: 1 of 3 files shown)


flashcode:
Thanks!

src/core/wee-crypto.c
```diff
@@ -524,6 +524,10 @@ weecrypto_totp_generate_internal (const char *secret, int length_secret,
     int rc, offset, length;
     unsigned long bin_code;
 
+#if __BYTE_ORDER == __BIG_ENDIAN
```
flashcode, line 527: Hi @luk1337,
I just noticed that the add of this condition make tests fail and procude invalid TOTP on FreeBSD (I'm testing with FreeBSD 14.0 on KVM/QMU with x86_64 arch).
If this condition is really needed on s390x, could it be changed to be more restrictive and apply only on this OS?

luk1337: https://github.com/weechat/weechat/pull/2171, idk if it will compile on all platforms tho, let's see.

PR #2022, opened 2023-09-26
core: replace inet_addr() with inet_pton()

luk1337:
man pages as well as rpminspect suggest that we shouldn't be using inet_addr().

### DIFF
--- a/src/core/wee-network.c
+++ b/src/core/wee-network.c
@@ -554,10 +554,12 @@ network_pass_socks4proxy (struct t_proxy *proxy, int sock, const char *address,
     socks4.method = 1;
     socks4.port = htons (port);
     network_resolve (address, ip_addr, NULL);
-    socks4.address = inet_addr (ip_addr);
     strncpy (socks4.user, username, sizeof (socks4.user) - 1);
 
     free (username);
+
+    if (inet_pton (AF_INET, ip_addr, &socks4.address) != 1)
+        return 0;
 
     length = 8 + strlen (socks4.user) + 1;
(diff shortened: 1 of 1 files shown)


flashcode (changes requested):

src/core/wee-network.c
```diff
@@ -554,7 +554,8 @@ network_pass_socks4proxy (struct t_proxy *proxy, int sock, const char *address,
     socks4.method = 1;
     socks4.port = htons (port);
     network_resolve (address, ip_addr, NULL);
-    socks4.address = inet_addr (ip_addr);
+    if (!inet_pton (AF_INET, ip_addr, &socks4.address))
```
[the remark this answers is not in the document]
luk1337: Fixed both.

src/core/wee-network.c
```diff
@@ -554,11 +554,13 @@ network_pass_socks4proxy (struct t_proxy *proxy, int sock, const char *address,
     socks4.method = 1;
     socks4.port = htons (port);
     network_resolve (address, ip_addr, NULL);
-    socks4.address = inet_addr (ip_addr);
     strncpy (socks4.user, username, sizeof (socks4.user) - 1);
 
     free (username);
 
+    if (inet_pton (AF_INET, ip_addr, &socks4.address) != 1)
+        return 1;
```
flashcode, line 562: Last change to do: return value must be 0 here (error).

flashcode (changes requested):

src/core/wee-network.c
luk1337: Done.

PR #2023, opened 2023-09-26
core, relay: make inet_pton() return value checks more strict

luk1337:
inet_pton() can return values < 0.


flashcode (changes requested):
Thanks, I just noticed there are two other calls to `inet_pton` to update in wee-network.c with a similar issue.

PR #2030, opened 2023-10-13
Improve error handling for hook_url

trygveaa:
This makes the callback being called when pthread_create fails, adds error codes to `output` for easier use in plugins/scripts and removes printing the errors to the core buffer unless `weechat_debug_core` is enabled. See the commit messages for details.

If you would like to keep the printing to core, we could add an option to the `option` hashtable to hook_url to disable it instead of disabling it by default.

### DIFF
--- a/src/core/hook/wee-hook-url.c
+++ b/src/core/hook/wee-hook-url.c
@@ -126,14 +126,22 @@ void *
 hook_url_transfer_thread (void *hook_pointer)
 {
     struct t_hook *hook;
+    int url_rc;
+    char str_error_code[12];
 
     hook = (struct t_hook *)hook_pointer;
 
     pthread_cleanup_push (&hook_url_thread_cleanup, hook);
 
-    (void) weeurl_download (HOOK_URL(hook, url),
-                            HOOK_URL(hook, options),
-                            HOOK_URL(hook, output));
+    url_rc = weeurl_download (HOOK_URL(hook, url),
+                              HOOK_URL(hook, options),
+                              HOOK_URL(hook, output));
+
+    if (url_rc != 0)
+    {
+        snprintf (str_error_code, sizeof (str_error_code), "%d", url_rc);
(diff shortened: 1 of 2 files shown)


src/core/hook/wee-hook-url.c
```diff
@@ -220,11 +231,25 @@ hook_url_transfer (struct t_hook *hook)
                          &hook_url_transfer_thread, hook);
     if (rc != 0)
     {
-        gui_chat_printf (NULL,
-                         _("%sError running thread in hook_url: %s (URL: \"%s\")"),
-                         gui_chat_prefix[GUI_CHAT_PREFIX_ERROR],
-                         strerror (rc),
-                         HOOK_URL(hook, url));
+        snprintf (str_error, sizeof (str_error),
+                  "error calling pthread_create (%d)", rc);
```
trygveaa, line 235: I don't know how to get a human readable string of the pthread_create error code.

trygveaa:
Need to add some documentation for the error codes too, if you accept this.

src/core/hook/wee-hook-url.c
trygveaa, line 235: Should this be translated? Related to this comment: https://github.com/weechat/weechat/commit/6468f7761f9a202bce29f4f958e783112d751fc8#r129887449

trygveaa:
I added error code 5 for pthread errors and 6 for timeout. I see hook_process uses -2 for both fork error and timeout, and I would like to have different error codes for pthread and timeout, so that's why I didn't use the same as hook_process.

src/core/hook/wee-hook-url.c
flashcode: I think it should not be translated as it's not for display, but an info sent to the callback (so it's easier to parse in a plugin/script).

flashcode:
Merged, thanks!

PR #2044, opened 2023-11-19
Use issue forms instead of templates

NetSysFire:
I briefly discussed this with @flashcode in #weechat some time ago and finally got to it.

To reiterate: Github has issue forms, which are in beta since 2 years but are still widely used and stable, which have their documentation there: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms

That means no more raw markdown editing and instead fancy text boxes and such.

I took the liberty to adjust some descriptions in a relatively minor way since the issue form does have labels and placeholders and such and more things needed to be written.

Testing:

Click "View File" on the added yaml files. Github will render them for you and you have a live preview.
If you merge this PR, the transition will be seamless and no further actions need to be done. Github detects it automagically.


flashcode:
Merged, thanks.

PR #2055, opened 2023-12-26
ruby: add detection for Ruby 3.3

faisal:
This PR makes it possible to build with Ruby 3.3, which the Ruby team just released.

This PR is based on https://github.com/weechat/weechat/pull/1865.


flashcode:
Thanks!

PR #2075, opened 2024-02-19
Refactor Tcl plugin

DasBrain:
simplifies things to the point where the plugin compiles with Tcl 9.0


flashcode:
Thanks, merged!

PR #2083, opened 2024-02-26
Refactor the Tcl plugin a bit more

DasBrain:
* Don't care about `Tcl_IsShared`.
* simplified constant initialization.


flashcode:
Thanks, commits squashed and merged.

PR #2089, opened 2024-03-09
GUI: fix potential integer overflows by casting to `size_t`

mario-campos:
Not sure how big `gui_color_term_colors` can realistically get, but it has type `int` so it's possible to overflow via the multiplication _before_ it is implicitly cast to a `size_t` function parameter.

By explicitly casting to a larger type (`size_t`) _before_ the multiplication, the compiler will implicitly promote the other operands to type `size_t`, avoiding the risk of overflow.

Source: [Chapter 4. Type Conversions, _C in a Nutshell_](https://www.oreilly.com/library/view/c-in-a/0596006977/ch04.html)


flashcode:
Thanks!

PR #2102, opened 2024-03-22
Add CodeQL static analysis

mario-campos:
[CodeQL](https://codeql.github.com/) is GitHub's enterprise static-analysis tool, which is free for public, open-source projects. This workflow will run CodeQL in PRs and on every push to the default "master" branch.

I've additionally created a configuration file that ignores a certain check, because, with it, CodeQL creates a lot of false positives.


.github/workflows/codeql.yml
```diff
@@ -0,0 +1,45 @@
...
+on:
+  push:
+    branches: [ "master" ]
+  pull_request:
+    branches: [ "master" ]
+  schedule:
+    - cron: '22 9 * * 2'
+
+env:
+  WEECHAT_DEPENDENCIES: devscripts equivs python3-pip cmake ninja-build lcov pkg-config libncurses-dev gem2deb libperl-dev python3-dev libaspell-dev liblua5.3-dev tcl8.6-dev guile-3.0-dev libv8-dev libcurl4-gnutls-dev libgcrypt20-dev libgnutls28-dev libzstd-dev zlib1g-dev curl libcpputest-dev php-dev libphp-embed libargon2-dev libsodium-dev flake8 pylint python3-bandit asciidoctor ruby-pygments.rb shellcheck
```
flashcode, line 12: Thanks for the PR, there's a codeql branch, I made an attempt 2 years ago but it was not working.

I wonder if these dependencies can be somewhat put in a file included in both ci.yml and codeql.yml?
I don't know if we can include yaml files here.

flashcode (changes requested):

.github/workflows/codeql.yml
mario-campos: >I wonder if these dependencies can be somewhat put in a file included in both ci.yml and codeql.yml?
I don't know if we can include yaml files here.

I'm sure there's a way; Actions is quite flexible.

But if the goal is to just have one dependency list, then perhaps the simplest solution would be to merge the codeql.yml and ci.yml file together into one file with 2 jobs? Then they could both reference the same environment variable.

mario-campos:
For the record, I explicitly disabled the `cpp/wrong-type-format-argument` query because it generated a lot of spurious results (see below images). And I found it too difficult to find the "signal" through the "noise." But, if you would prefer to keep those results (so that you may with them what you wish), let me know -- happy to re-enable that query.

---

Screenshot 2024-03-27 at 10 03 38 AM

---

Screenshot 2024-03-27 at 10 04 09 AM

---

Screenshot 2024-03-27 at 10 04 26 AM

flashcode:
@mario-campos: I think it's better to include these warnings as well, I'll fix in code to mute them (or fix in case of real issues).

Could you please merge the new file in ci.yml, as a new codeql job at the end?
So we have just one environment variable with dependencies.

Thanks for your work!

flashcode:
Thanks for changes.

Last question: why scheduling a cron for CI and CodeQL?
Isn't it enough to run on any push and pull request?

mario-campos:
It's a bit of a myth that if the code doesn't change, the results won't&mdash;it might still find new vulns. So it's not required, but it's recommended to periodically scan the code during the periods of inactivity. Your choice.

flashcode:
OK, the project is active, but let's keep it, it doesn't hurt.

flashcode (approved):

flashcode:
Thanks, I squashed commits and merged.

PR #2103, opened 2024-03-22
relay: set server last_client_disconnect only for successful connections

Leo3418:
With certain option settings, a relay client receives the backlog since the last_client_disconnect timestamp, which ideally would let the user conveniently pick up messages where they left off.

However, if an unsuccessful connection (e.g. failed authentication, port scan, or other spam request) to the relay was made before the user successfully reconnects to it, the user would fail to see messages received before the unsuccessful connection in the relay client, hence they might miss messages arrived while they were away.  This is because last_client_disconnect is also updated when an unsuccessful connection closes.

To solve this, update last_client_disconnect only when a *successfully-connected* client is disconnected, so unsuccessful connections cannot affect it.

### DIFF
--- a/src/plugins/relay/relay-client.c
+++ b/src/plugins/relay/relay-client.c
@@ -1426,6 +1426,7 @@ relay_client_new (int sock, const char *address, struct t_relay_server *server)
                                       address : "local");
         new_client->real_ip = NULL;
         new_client->status = RELAY_STATUS_CONNECTING;
+        new_client->ever_connected = 0;
         new_client->protocol = server->protocol;
         new_client->protocol_string = (server->protocol_string) ? strdup (server->protocol_string) : NULL;
         new_client->protocol_args = (server->protocol_args) ? strdup (server->protocol_args) : NULL;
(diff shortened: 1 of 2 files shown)


flashcode:
Hi,

Thanks for the PR.

I think it's not needed to store any state in the client itself, when the status is set to an "end" status, we could just check if the status of the client was "RELAY_STATUS_CONNECTED" before ending, in which case only we can set the "last_client_disconnect" time.

Leo3418:
> Hi,
> 
> Thanks for the PR.
> 
> I think it's not needed to store any state in the client itself, when the status is set to an "end" status, we could just check if the status of the client was "RELAY_STATUS_CONNECTED" before ending, in which case only we can set the "last_client_disconnect" time.

Thanks for your feedback!  I will look into implementing your suggestion and update this pull request later.

flashcode:
For your information, I fixed the issue in commit 2f5936bc9f8e297792a663d153267d1a6821107b, so I closed this PR.

PR #2106, opened 2024-04-08
Add support for $XDG_STATE_HOME

### LINKED ISSUE #1747: Add support for $XDG_STATE_HOME
## Feature description

Background: Support for the XDG base directory spec was recently closed in #1285. 

Also recently, freedesktop.org has [extended the spec](https://specifications.freedesktop.org/basedir-spec/latest/ar01s03.html) to add support for `$XDG_STATE_HOME`:

> $XDG_STATE_HOME defines the base directory relative to which user-specific state files should be stored. If $XDG_STATE_HOME is either not set or empty, a default equal to $HOME/.local/state should be used.
>
> The $XDG_STATE_HOME contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in $XDG_DATA_HOME. It may contain:
>
> * actions history (logs, history, recently used files, …)
>
>  * current state of the application that can be reused on a restart (view, layout, open files, undo history, …)

62832:
The most recent version of the [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) from 2021 introduces the new `$XDG_STATE_HOME` directory for non-essential "state" data such as history and logs.

> The `$XDG_STATE_HOME` contains state data that should persist between (application) restarts, but that is not important or portable enough to the user that it should be stored in `$XDG_DATA_HOME`. It may contain:
> - actions history (logs, history, recently used files, …)
> - current state of the application that can be reused on a restart (view, layout, open files, undo history, …)

This PR adds support for this directory via the `weechat_state_dir` property, and uses this directory by default from now on to store log files. Closes #1747.


flashcode:
Hi,

Thanks for this PR.

There's a discussion about what to move to the XDG state directory in this issue: #1747.
And I concur with @trygveaa about the fact that buffer log files (from logger plugin) are important and just not "state", and should stay in data directory.

The specification is not very clear about what should be moved anyway.

Any thoughts from other people?

flashcode:
By the way I'm not even sure if the weechat.log file should be in $XDG_STATE_HOME or $XDG_RUNTIME_DIR directory, as it's a temporary log file used for lock (it's used to prevent two weechat instances to use the same home directory).

62832:
Considering that `weechat.log` doesn't remove itself upon exit like a regular "lock" file would, I would treat it as a regular log file for `$XDG_STATE_HOME`.

flashcode:
Thanks, I squashed commits and merged without the changes in logger plugin.

PR #2116, opened 2024-05-26
tests: relay: add missing <string.h> include

luk1337:
Fixes build error on Fedora 40.

### DIFF
--- a/tests/unit/plugins/relay/api/remote/test-relay-remote-network.cpp
+++ b/tests/unit/plugins/relay/api/remote/test-relay-remote-network.cpp
@@ -25,6 +25,7 @@
 
 extern "C"
 {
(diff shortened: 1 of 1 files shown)


flashcode:
Thanks for the fix.

For my information, do you know why it fails to compile on Fedora while it compiles fine on Debian (any version), with gcc or clang?

This is not the first time it happens, and currently I have no known way to check that such header is missing.

luk1337:
> Thanks for the fix.
> 
> For my information, do you know why it fails to compile on Fedora while it compiles fine on Debian (any version), with gcc or clang?
> 
> This is not the first time it happens, and currently I have no known way to check that such header is missing.

No idea.

PR #2118, opened 2024-05-26
tests: relay: fix HotlistToJson test on s390x

luk1337:
```
1: error: Failure in TEST(RelayApiMsg, HotlistToJson)
1: 	expected <2024-05-26T10:29:37.716512Z>
1: 	but was  <2024-05-26T10:29:37.000000Z>
1: 	difference starts at position 20 at: <T10:29:37.000000Z   >
```


flashcode:
Thanks, I merged with a minor change: not using the weechat plugin header, but using directly hdata functions (in tests all core functions are directly used, even in plugin tests).

PR #2119, opened 2024-05-26
tcl: fix return value for hdata_longlong

luk1337:
Using API_RETURN_LONG() here breaks tests on i368.


flashcode:
Thanks!

I'll fix a similar issue in php plugin, I've a failure on 32-bit GNU/Hurd.

PR #2123, opened 2024-05-30
core: always send input_text_changed signal on buffer set input

trygveaa:
When the buffer input is set to the same string it already has, we should still run the input modifiers and send the input_text_changed signal. This is because plugins/scripts may rely on them being run when they set the input.

An example is the go.py script. When you start it, it stores the input and sets it to an empty string, and uses some input modifiers to display custom data in the input field. Since this function just returned early if the input was empty, it had a bug where if you had /go bound to a key and pressed the key when you had an empty input, the buffers go adds to the input wouldn't be shown until you press another key.

This fixes a regression that was introduced in commit 7addd1bf0070c8b6c36f6e1d90245c44f742d7ee.

### DIFF
--- a/src/gui/gui-buffer.c
+++ b/src/gui/gui-buffer.c
@@ -2353,7 +2353,7 @@ gui_buffer_set_input_prompt (struct t_gui_buffer *buffer,
 void
 gui_buffer_set_input (struct t_gui_buffer *buffer, const char *input)
 {
-    if (!buffer || (string_strcmp (buffer->input_buffer, input) == 0))
+    if (!buffer)
         return;
 
     gui_buffer_undo_snap (buffer);


src/gui/gui-buffer.c
```diff
@@ -2353,7 +2353,7 @@ gui_buffer_set_input_prompt (struct t_gui_buffer *buffer,
 void
 gui_buffer_set_input (struct t_gui_buffer *buffer, const char *input)
 {
-    if (!buffer || (string_strcmp (buffer->input_buffer, input) == 0))
```
trygveaa, line 2356: I considered moving the condition around the `gui_buffer_undo_snap` and `gui_input_replace_input` lines below, but there doesn't seem to be any point really. An undo point will only be added if the input actually changes.

flashcode:
Instead I updated the script `go.py` to force a refresh of bar item `input_text` as soon as the modifier hook is added.

I'm closing this PR without merging, as the current behavior is good: the signal "input_text_changed" is sent only if the input actually changes, which makes sense.

PR #2132, opened 2024-06-07
tests: relay: fix relay_http_parse_header function prototype

luk1337:


flashcode:
Good catch, thanks!

I've backported on branch 4.3 as well.

PR #2136, opened 2024-06-16
src/plugins/weechat-plugin: fix muloveflow in weechat_va_format

mikoxyz:
this is done by refactoring the entire macro to be almost the same as string_asprintf


flashcode:
Thanks, merged with minor style changes.

PR #2143, opened 2024-07-03
irc: fix lag check interval comment

elalonde:
The network lag check interval is stored in irc_config_network_lag_check, which does not default to 5 seconds.

Looks like the erroneous comment was introduced in 9d740130365a7aecd34f64a5c502eff74bf16d3e


flashcode:
Thanks!

PR #2171, opened 2024-08-14
core: Replace manual endianness swap with htobe64() call

luk1337:
https://github.com/weechat/weechat/pull/2021#discussion_r1716381589

### DIFF
--- a/src/core/core-crypto.c
+++ b/src/core/core-crypto.c
@@ -38,6 +38,10 @@
 #include "core-string.h"
 #include "../plugins/plugin.h"
 
+#ifdef htonll
+#define htobe64 htonll
+#endif
+
 char *weecrypto_hash_algo_string[] = {
     "crc32",
     "md5",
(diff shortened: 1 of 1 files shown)


flashcode:
It seems you need to add `#include <endian.h>`, this could possibly fix compilation on macOS.

luk1337:
@flashcode can you restart macos build? It failed for no apparent reason.

flashcode:
Done.

src/core/core-crypto.c
```diff
@@ -23,6 +23,10 @@
 #include "config.h"
 #endif
 
+#if defined(__APPLE__)
```
luk1337, line 26: should I move it above config.h include?

flashcode:
Also from the man page:

```
Similar  functions are present on the BSDs, where the required header file is
<sys/endian.h> instead of <endian.h>.  Unfortunately, NetBSD, FreeBSD, and glibc haven't
followed the original OpenBSD naming convention for these functions, whereby the nn
component always appears at the end of the function name (thus, for example, in NetBSD,
FreeBSD, and glibc, the equivalent of OpenBSDs "betoh32" is "be32toh").
```

So it could fail on FreeBSD, I'm going to test.

Also I think including the header should be done as well, I'm pretty sure this can fail on some OS/distributions without it.

src/core/core-crypto.c
flashcode: I think it could be moved below, after all includes.
No other choice than defining this on macOS?

luk1337:
> Also from the man page:
> 
> ```
> Similar  functions are present on the BSDs, where the required header file is
> <sys/endian.h> instead of <endian.h>.  Unfortunately, NetBSD, FreeBSD, and glibc haven't
> followed the original OpenBSD naming convention for these functions, whereby the nn
> component always appears at the end of the function name (thus, for example, in NetBSD,
> FreeBSD, and glibc, the equivalent of OpenBSDs "betoh32" is "be32toh").
> ```
> 
> So it could fail on FreeBSD, I'm going to test.
> 
> Also I think including the header should be done as well, I'm pretty sure this can fail on some OS/distributions without it.

https://man.freebsd.org/cgi/man.cgi?query=byteorder&sektion=9 has htobe64(), so should be ok?

Also, sys/endian.h doesn't exist on macos ;/

src/core/core-crypto.c
luk1337: did that:

```
#ifdef htonll
#define htobe64 htonll
#endif
```

flashcode:
It seems complicated to use these functions while being portable…
But I'm sure it's doable, I'll then make tests on Linux, FreeBSD, GNU/Hurd, Cygwin, Windows WSL.

flashcode:
It's OK on FreeBSD, TOTP are valid, tests are OK.

flashcode:
Thanks for the quick fix, merged (branches master and 4.3).

PR #2173, opened 2024-08-17
lua: fix broken LUA_VERSION check

luk1337:
/usr/include/lua-5.1/lua.h:19:25: error: token ""Lua 5.1"" is not valid in preprocessor expressions

### DIFF
--- a/src/plugins/lua/weechat-lua.c
+++ b/src/plugins/lua/weechat-lua.c
@@ -1272,7 +1272,7 @@ weechat_plugin_init (struct t_weechat_plugin *plugin, int argc, char *argv[])
 #if defined(LUA_VERSION_MAJOR) && defined(LUA_VERSION_MINOR)
     weechat_hashtable_set (plugin->variables, "interpreter_version",
                            LUA_VERSION_MAJOR "." LUA_VERSION_MINOR);
-#elif LUA_VERSION
+#elif defined(LUA_VERSION)
     weechat_hashtable_set (plugin->variables, "interpreter_version",
                            LUA_VERSION);
 #else


flashcode:
Of course there's an issue there, but out of curiosity, on which OS or with which compiler does this fail like this?
Because the CI was green and manual tests OK on several OS / compilers.

luk1337:
> Of course there's an issue there, but out of curiosity, on which OS or with which compiler does this fail like this? Because the CI was green and manual tests OK on several OS / compilers.

Fedora.

flashcode:
Hmm then I wonder what's different in Fedora that causes this fail.
If ever you know, it would be interesting for me to catch such error in CI, either by testing with Fedora (if no other choice), either by setting the CI in the same condition/environment to reproduce such issue.

flashcode:
Anyway, I'll release a version 4.4.1 now to fix the issue, if that's the only one you had.

luk1337:
> Anyway, I'll release a version 4.4.1 now to fix the issue, if that's the only one you had.

yeah, that was 4.4.0 issue

flashcode:
Which version of Fedora and compiler?
I think it could be related to the compiler version.

luk1337:
> Which version of Fedora and compiler? I think it could be related to the compiler version.

40, but I found out why it was happening, I'll open PR shortly.

luk1337:
> Hmm then I wonder what's different in Fedora that causes this fail. If ever you know, it would be interesting for me to catch such error in CI, either by testing with Fedora (if no other choice), either by setting the CI in the same condition/environment to reproduce such issue.

this issue in particular probably wouldn't be reproduced by CI, as this is just my local setup quirk.

flashcode:
OK, so I'll wait before releasing 4.4.1, especially with the other PR, I want to be sure there are no regressions anywhere.

oxzi:
Just for your information, I experienced the same build error on a current OpenBSD. Applying the patch worked for me as well.

flashcode:
Thanks, I'll release 4.4.1 in a few hours, with this fix.

flashcode:
Version 4.4.1 has been released, and in addition CI running on Rocky Linux 9 has been added in master branch: 452a20c05e36a750605bdfa67fbeb6dba03754ae.

PR #2174, opened 2024-08-17
cmake: find 'lua' first

luk1337:
On Fedora, `lua` is an up to date package.


luk1337:
Prior to this change, it was using https://src.fedoraproject.org/rpms/compat-lua, which is lua 5.1.5 with a few patches.

Issue #2178, opened 2024-09-03
Integer Overflow to Buffer Overflow vulnerability in "string_free_split" functions


Crispy-fried-chicken:
### Question

Hi,
   we have detected that your project may be vulnerable to Integer Overflow to Buffer Overflow in the function of `string_free_split_shared `, `string_free_split`, `string_free_split_command` and `string_free_split_tags`  in the file of `src/core/core-string.c` . It shares similarities to a recent CVE disclosure [CVE-2024-6381](https://nvd.nist.gov/vuln/detail/CVE-2024-6381) in the [mongo-c-driver](https://github.com/mongodb/mongo-c-driver).

**The source vulnerability information is as follows:**

> Vulnerability Detail:
> CVE Identifier: CVE-2024-6381
> Description: The bson_strfreev function in the MongoDB C driver library may be susceptible to an integer overflow where the function will try to free memory at a negative offset. This may result in memory corruption. This issue affected libbson versions prior to 1.26.2
> Reference: https://nvd.nist.gov/vuln/detail/CVE-2024-6381
> Patch: https://github.com/mongodb/mongo-c-driver/commit/effd95c34ad421df94eec7c69236f0e4172552d0

Would you help to check if this bug is true? If it's true, I'd like to open a PR for that if necessary. Thank you for your effort and patience!

flashcode:
Hi,

Thanks for the issue, I confirm there can be an integer overflow here, and at least these functions must be fixed (I'm checking if other functions could be affected as well).

Did you find this manually or automatically with a tool? If yes, which one?
I make regular scans of the source code for such issues, and it was not reported.

This issue should not really happen in practice, because INT_MAX is 2,147,483,647, so the integer overflow happens with this number of split items, that means with the separator between each item, the source string was at least 4,294,967,294 bytes long.

**Important note:** for any security issue found in WeeChat (even if you're not sure it's one), please do not post in the public tracker, see file [CONTRIBUTING.md](https://github.com/weechat/weechat/blob/master/CONTRIBUTING.md#security-reports).
This allows time to properly fix the issue in all versions, then announce the vulnerability publicly. Thanks!

flashcode:
No e-mail needed, we can go on here, as this has a minor impact on users (should not happen in practice).

Yes, a fix is welcome.
I see in mongodb they just use a pointer without index variable (https://github.com/mongodb/mongo-c-driver/pull/1552), which is fine to me.

flashcode:
[...]

PR #2179, opened 2024-09-03
fix https://github.com/weechat/weechat/issues/2178

Crispy-fried-chicken:


flashcode:
Merged with a bug fixed in function string_free_split_tags (pointer type was wrong) and some changes on style.

Thanks!

PR #2180, opened 2024-09-09
core: Fix build on Android to define htobe64

fornwall:
This is needed to get `htobe64` defined on Android - see https://android.googlesource.com/platform/bionic/+/refs/heads/main/libc/include/sys/endian.h.


flashcode:
Thanks!

By the way, do you know if it's easy to test the build on Androind in GitHub Actions, to quickyl detect such problem in the CI?

PR #2183, opened 2024-09-11
cmake: require 3.12, resolve circular dependency

evelikov:
In order to resolve the circular dependency, we need to annotate the
respective static libraries as object "libraries."

This requires cmake 3.12, where Debian 10 (old old stable) and Ubuntu
20.04 have 3.13 and 3.16 respectively.

An alternative is to manually track and use the respective source files.
Since all platforms of interest meet the cmake requirement, we can opt
for the simpler approach.

---

What tends to be the policy wrt minimum required version - is Debian/Ubuntu the most significant factor? Are there any objections, if I bump a few other projects and do some cleanups?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -242,8 +242,6 @@ endif()
 # Check for CURL
 find_package(CURL REQUIRED)
 
-set(STATIC_LIBS weechat_plugins)
-
 find_library(DL_LIBRARY
   NAMES dl
   PATHS /lib /usr/lib /usr/libexec /usr/local/lib /usr/local/libexec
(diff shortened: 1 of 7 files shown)


flashcode:
Thanks, merged!

flashcode:
Requiring CMake 3.12 is OK, if that was your question.

flashcode:
Yes we could drop support for very old libraries, providing that:

* almost all users/distros still don't use them any more; think about old Ubuntu still supported with extended LTS support (for example: Ubuntu 16.04 LTS (Xenial Xerus) is still supported until April 2026, but anyway on these distros users are not expected to compile latest WeeChat versions, but only apply patches of old released versions)
* it removes complex code (eg: removing just one localized #ifdef with just one line of code is not worth removing this support, except it there's another good reason to do it)

If you plan to work on this, please make a separate PR for each library, so that it's easier to integrate.

PR #2186, opened 2024-09-11
Spelling

jsoref:
Fixes misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling).

The misspellings have been reported at https://github.com/jsoref/weechat/actions/runs/10818633560#summary-30014686649

The action will report that the changes in this PR would make it happy: https://github.com/jsoref/weechat/actions/runs/10818633766


doc/en/weechat_dev.en.adoc
```diff
@@ -1362,5 +1362,5 @@ warnings, ... These words must be kept unchanged:
 When there is a name after `+<<link_name>>+`, then you must translate it:
```
jsoref, line 1362: I know that `must` is used in the instructions, but it's fairly unhelpful.

There will be translations where a thing is translated to itself and it's meaningless to insist that it must be translated. It makes even less sense later.

src/plugins/alias/alias-command.c
```diff
@@ -353,7 +353,7 @@ alias_command_init ()
     weechat_hook_command (
         "alias",
         N_("list, add or remove command aliases"),
-        /* TRANSLATORS: only text between angle brackets (eg: "<name>") must be translated */
+        /* TRANSLATORS: only text between angle brackets (eg: "<name>") should be translated */
```
jsoref, line 356: here the text is even more problematic, "**only**...**must**" implies "other things **may**".

src/plugins/alias/alias.c
```diff
@@ -428,7 +428,7 @@ alias_hook_command (struct t_alias *alias)
 
     /*
      * build string with priority and name: the alias priority is 2000, which
-     * is higher than default one (1000), so the alias is executed before a
+     * is greater than default one (1000), so the alias is executed before a
```
jsoref, line 431: often `greater than` is more appropriate than `higher than`, here I'm on the fence.

flashcode:
Merged, thanks!

I made a few changes:

- fix similar typos in other files (for example in docs in other languages where part of text is not yet translated and still in English)
- remove your changes in weechat.pot (this file is auto-generated); instead I have regenerated and updated translations on top of your commits
- change sentence: `/* TRANSLATORS: only text between angle brackets (eg: "<name>") should be translated */` to `/* TRANSLATORS: only text between angle brackets (eg: "<name>") may be translated */`.

PR #2193, opened 2024-09-29
Bump to gnutls 3.3.0, use pkg_check_modules()

evelikov:


flashcode:
Merged, thanks!

flashcode:
Yes I just noticed the macOS CI is failing, `libintl.h` is not found.
I think this is not related to translations being removed, but this header file was indirectly found with the GnuTLS includes present in CMake files before your changes.

I'll check why this file is not found on macOS.

PR #2194, opened 2024-09-29
core: remove ifdef checks for core components

evelikov:
The gcrypt, gnutls, libcurl and zlib are core libraries/components that we always build against. Remove the #ifdef checks - the symbols must be available at build.


flashcode:
Hi,

I know the libs are required, but if I remember well, the constants checked were not available in old versions of the libs, hence these checks to display an unknown version in this case.

So we should check that all these libs provide the constants in all versions still supported.

PR #2195, opened 2024-09-29
Fix check_curl_symbols.py, bump curl to 4.47.0

evelikov:
Seems like our regex parsing in check_curl_symbols.py is a little broken ;-) Fix that, bump curl and update the script to omit the old version checks.

See the individual commits for more details.


flashcode:
Hi,
In your first commit message, you mean Curl versions 7.47.0 and 7.64.0 (and not 4.47.0 and 4.64.0), right?

flashcode:
Also the CI failed because of two flake8 errors in script `check_curl_symbols.py`.

flashcode:
Thanks!

PR #2201, opened 2024-10-19
core: simplify plugin callback by refactoring if statements

JamesCMorey:
**Refactor to Simplify Control Flow**

This refactor removes unnecessary nesting to make the control flow more readable.

**Current Control Flow:**

There are 3 possible paths through this code:

**Path 1**: argc > 3 (which implies argc > 2)
**Path 2**: argc == 3
**Path 3**: argc <= 2

**Proposed Changes:**

The nested if statements are unraveled, and the logic is simplified by handling these cases up front.

- There's no need to check argc > 2 and argc > 3 separately, as argc > 3 implies argc > 2.
- For 3 >= argc > 2, argc can only be equal to 3, so this is checked explicitly.
- The final case (argc <= 2) is handled in the else clause, as in the original.

### DIFF
--- a/src/core/core-command.c
+++ b/src/core/core-command.c
@@ -5189,30 +5189,27 @@ COMMAND_CALLBACK(plugin)
 
     if (string_strcmp (argv[1], "reload") == 0)
     {
-        if (argc > 2)
+        if (argc > 3)
         {
-            if (argc > 3)
-            {
-                plugin_argv = string_split (
+            plugin_argv = string_split (
                     argv_eol[3], " ", NULL,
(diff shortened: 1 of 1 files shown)


JamesCMorey:
Closing PR to follow best practice of using a feature branch. Will resubmit PR from new branch shortly.

flashcode:
Merging to branch `main` is OK.

I think the change could make even more sense by checking arguments, from 2 to 3+, like this:

```C
if (argc == 2)
{
    // ...
}
else if (argc == 3)
{
    // ...
}
else
{
    // ...
}
```

PR #2204, opened 2024-10-30
tests: migrate away from removed ast features

luk1337:
See https://github.com/python/cpython/pull/119563.


luk1337:
I have no idea if this is correct, but it passes the tests

flashcode:
Thanks!

PR #2205, opened 2024-10-30
relay/api: support passing auth in sub protocol header

trygveaa:
The API for connecting to WebSockets in browsers unfortunately doesn't support setting any Authorization header. This means that before this commit it was impossible to connect to the API relay from a web browser. The only thing that can be set apart from the URL is the Sec-WebSocket-Protocol header. Therefore this allows you to send the auth token in this header.

This is a weird way to send auth, but it seems to be the best one that makes it possible for browsers to connect. Kubernetes also does it this way: https://github.com/kubernetes/kubernetes/pull/47740

Here is a post describing the different ways to make it possible for a browser to authenticate against a websocket connection, and it also recommends doing it this way:
https://stackoverflow.com/questions/4361173/http-headers-in-websockets-client-api/77060459#77060459

Note that when this header is used to pass auth, the client also needs to specify the `api.weechat` sub protocol. This is because the client and server have to agree on a sub protocol when this header is specified, and in order to not send the fake protocol used for auth back to the client, we require specifying the protocol `api.weechat`, which the server then returns to the client. This is only necessary when the Sec-WebSocket-Protocol header is used. If the Authorization header is used for auth as before, nothing changes.

### DIFF
--- a/doc/en/weechat_relay_api.en.adoc
+++ b/doc/en/weechat_relay_api.en.adoc
@@ -101,7 +101,9 @@ Examples:
 [[authentication]]
 == Authentication
 
-The password must be sent in the header `Authorization` with `Basic` authentication schema.
+The password must be sent in the header `Authorization` with `Basic`
+authentication schema or in the header `Sec-WebSocket-Protocol` (see details
+below).
 
 The password can be sent as plain text or hashed, with one of these formats
 for user and password:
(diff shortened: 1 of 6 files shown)


trygveaa:
Should write about this in the docs. I can do it if you agree with supporting this way of passing auth.

flashcode:
Yes it's OK for me so you can update docs.

Could you please also update tests on the modified functions to test the new way to authenticate?

Thanks!

trygveaa:
I added some tests and documentation now.

flashcode:
Thanks, merged!

I fixed a typo: `protool_array` → `protocol_array`.

PR #2206, opened 2024-10-30
relay/api: combine request headers with the same name

trygveaa:
If a request repeats the same header name multiple times, merge the header values into a comma separated string. Previously, only the last header specified would be used.

For header fields that are defined as a comma-separated list, a client may choose to send it as multiple headers instead of one header with comma-separated values. The specification says that these are equivalent, so we can therefore join the headers into a comma-separated string.

This is specified at https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2 which says:

    A sender MUST NOT generate multiple header fields with the same field
    name in a message unless either the entire field value for that
    header field is defined as a comma-separated list [i.e., #(values)]
    or the header field is a well-known exception (as noted below).

    A recipient MAY combine multiple header fields with the same field
    name into one "field-name: field-value" pair, without changing the
    semantics of the message, by appending each subsequent field value to
    the combined field value in order, separated by a comma.  The order
    in which header fields with the same field name are received is
    therefore significant to the interpretation of the combined field
    value; a proxy MUST NOT change the order of these field values when
    forwarding a message.

### DIFF
--- a/src/plugins/relay/relay-http.c
+++ b/src/plugins/relay/relay-http.c
@@ -385,7 +385,7 @@ relay_http_parse_header (struct t_relay_http_request *request,
                          int ws_deflate_allowed)
 {
     char *pos, *name, *name_lower, *error, **items;
-    const char *ptr_value;
+    const char *existing_value, *ptr_value;
     int i, num_items;
     long number;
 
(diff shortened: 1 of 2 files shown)


trygveaa:
Yes, I added a test now.

flashcode:
Thanks, merged!

I just replaced the call to `weechat_string_concat` with the recommended macro `WEECHAT_STR_CONCAT` (exactly same result).

PR #2207, opened 2024-11-02
Add /api/completion route

nilscc:
Added route handler for `/api/completion`. Open tasks:

* Update documentation
* Update API specification yaml file

Let's first discuss the implementation. 🙂


src/plugins/relay/api/relay-api-protocol.c
```diff
@@ -663,6 +663,126 @@ RELAY_API_PROTOCOL_CALLBACK(buffers)
...
+        if (cJSON_IsNumber(json_buffer_id))
+        {
+            snprintf(str_id, sizeof(str_id),
+                     "%lld", (long long)cJSON_GetNumberValue(json_buffer_id));
+            ptr_buffer = weechat_buffer_search("==id", str_id);
+            if (!ptr_buffer)
+            {
+                relay_api_msg_send_error_json(
+                    client,
+                    RELAY_HTTP_400_BAD_REQUEST, NULL,
```
nilscc, line 702: 400 seems like the correct response, as the route is available (i.e. not 404), but the buffer ID is invalid. this is inconsistent with the handling of the `/api/input` handler (line 849+) though => change input route as well?

nilscc: Done.

tests/unit/plugins/relay/api/test-relay-api-msg.cpp
```diff
@@ -578,3 +578,20 @@ TEST(RelayApiMsg, HotlistToJson)
...
 }
+
+/*
+ * Tests functions:
+ *   relay_api_msg_completion_to_json
+ */
+
+TEST(RelayApiMsg, CompletionToJson)
+{
+    cJSON *json; //, *json_obj, *json_count;
```
flashcode, line 589: It seems this test is not complete, and some commented variables are still there.
Please make a complete test by building a completion JSON object with a real completion.

You should be able to simulate a completion from user on core buffer with such code:

```C
gui_buffer_set (gui_buffers, "input", "/co"); 
gui_buffer_set (gui_buffers, "input_pos", "3");
gui_input_complete_next (gui_buffers);
```

flashcode (changes requested):

flashcode:
Thanks, I merged with some changes:

- code style changes
- add completion context "null" (should happen rarely, but can be returned by the API) and defaults to "auto" if unknown (same as the "weechat" protocol)
- rename body field "data" to "command" (to be consistent with `POST /api/input`)
- return `RELAY_API_PROTOCOL_RC_MEMORY` when the completion can not be created (instead of "bad request")
- update OpenAPI spec YAML document (file `weechat-relay-api.yaml`)
- bump relay API version
- add missing include of header and fix unit tests

PR #2211, opened 2024-11-15
Update Serbian translation

eevan78:
Here is an update of Serbian translation of messages and documentation.


flashcode:
Thanks!

PR #2213, opened 2024-11-25
Only set Perl locale if the locale is broken

trygveaa:
It turns out that Debian has reverted the commit in Perl that broke the locale in their 5.38 branch, so it did not have the issue. However, the workaround we added to fix the locale apparently makes the version Debian/Ubuntu has crash on perl_destruct. I'm not sure why it makes it crash, but since it doesn't crash on newer Perl versions, I'm assuming that it's another bug with the locale handling in that Perl version.

To avoid the crash, make sure to only set the locale if we detect that it has been broken by Perl. We do this by checking if the value returned by wcwidth (160) (the first non-ascii printable character) has changed. If this value is not the same after the call to perl_construct, the locale has been broken.

I moved the call to Perl_setlocale to right after perl_construct, as the call to perl_construct is what breaks the locale.

I also changed the version check to match the versions that upstream include the commit that breaks the locale, and doesn't include the commit that fixes it.

### DIFF
--- a/src/plugins/perl/weechat-perl.c
+++ b/src/plugins/perl/weechat-perl.c
@@ -507,7 +507,9 @@ weechat_perl_load (const char *filename, const char *code)
     struct stat buf;
     char *perl_code;
     int length;
-#ifndef MULTIPLICITY
+#ifdef MULTIPLICITY
+    int wcwidth160;
+#else
     char pkgname[64];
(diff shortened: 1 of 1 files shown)


flashcode:
Thanks, merged with some changes:

- add missing include of `wchar.h`
- remove condition on Perl version, so we always fix the locale if needed

PR #2214, opened 2024-11-25
script: don't try to display the old line if it's NULL

trygveaa:
This fixes a crash which would happen if you scrolled the script buffer and then did a search which got fewer search results than the index of the selected line before the search. E.g. press page down to go to the second page and then search for `test`.


flashcode:
I fixed a similar issue on `/list` buffer (IRC): 423f609828884a8b363e33433b6a03103c214865.
I'll check if there can be similar issues with other buffers of this type, like `/fset`.

flashcode:
Thanks, merged.

Issue #2218, opened 2024-12-18
4.5.0 fails to link: `Undefined symbols: "_getcurx"` etc.


barracuda156:
### Describe the bug

Build fails at linking:
```
:info:build Undefined symbols:
:info:build   "_getcurx", referenced from:
:info:build       _gui_bar_window_print_string in libweechat_gui_curses_normal.a(gui-curses-bar-window.c.o)
:info:build   "_getcury", referenced from:
:info:build       _gui_bar_window_print_string in libweechat_gui_curses_normal.a(gui-curses-bar-window.c.o)
:info:build   "_getmaxx", referenced from:
:info:build       _gui_window_read_terminal_size in libweechat_gui_curses_normal.a(gui-curses-window.c.o)
:info:build   "_getmaxy", referenced from:
:info:build       _gui_window_read_terminal_size in libweechat_gui_curses_normal.a(gui-curses-window.c.o)
:info:build ld: symbol(s) not found
```

### Steps to reproduce

Run the build.

### Current behavior

```
[100%] Building C object src/gui/curses/normal/CMakeFiles/weechat.dir/main.c.o
[...]

PR #2221, opened 2024-12-21
CMakeLists: fix linking to curl and ncurses on macOS

barracuda156:
Closes: https://github.com/weechat/weechat/issues/2218

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -247,8 +247,10 @@ find_library(DL_LIBRARY
   PATHS /lib /usr/lib /usr/libexec /usr/local/lib /usr/local/libexec
 )
 if(DL_LIBRARY)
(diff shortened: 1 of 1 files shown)


barracuda156:
@flashcode This should fix the issue. Since I have no idea whether explicit flag is needed elsewhere, a safe way would be to avoid that on macOS.

flashcode:
Hi,

Thanks for the fix.

I just pushed an alternate way to fix the issue, that is not specific to macOS and should fix the detection of dl library everywhere.
Could you please try to build this branch: [fix-detection-dl-library](/weechat/weechat/tree/fix-detection-dl-library)?

barracuda156:
@ flashcode Yes, it also works. I will close this then and update a PR to MacPorts with your patch.

Issue #2046, opened 2023-11-21
weechat crashes on exit when both title.py and autosort.py are loaded (voidlinux)


laemeiqu:
## Bug summary
When using weechat-4.1.1_1 and weechat-python-4.1.1_1 under voidlinux. When scripts title.py and autosort.py are both installed weechat crashes on exit.

## Steps to reproduce

1. /script load title.py 
2. /script load autosort.py
3. /exit

If no crash at this time then reloading weechat and /exit will definitely crash weechat in voidlinux.

## Current behavior

Weechat crashes during exit when both scripts are installed.

## Expected behavior

Weechat doesn't crash on exit when both scripts are installed.

## Additional information

After bulding debug packages in voidlinux

gdb /usr/bin/weechat

Program received signal SIGSEGV, Segmentation fault.
                                                                             0x00007ffff626f76d in PyObject_GC_UnTrack () from /usr/lib/libpython3.12.so.1.0

bt full
```
#0  0x00007ffff626f76d in PyObject_GC_UnTrack () from /usr/lib/libpython3.12.so.1.0
No symbol table info available.
#1  0x00007ffff62a2099 in ?? () from /usr/lib/libpython3.12.so.1.0
No symbol table info available.
#2  0x00007ffff62251e5 in ?? () from /usr/lib/libpython3.12.so.1.0
No symbol table info available.
#3  0x00007ffff6220276 in ?? () from /usr/lib/libpython3.12.so.1.0
No symbol table info available.
#4  0x00007ffff6225bd5 in _PyModule_ClearDict () from /usr/lib/libpython3.12.so.1.0
No symbol table info available.
#5  0x00007ffff62eaaf6 in ?? () from /usr/lib/libpython3.12.so.1.0
No symbol table info available.
#6  0x00007ffff62ead6b in Py_EndInterpreter () from /usr/lib/libpython3.12.so.1.0
No symbol table info available.
#7  0x00007ffff66ea2a0 in weechat_python_unload (script=0x555555df3550)
    at /builddir/weechat-4.1.1/src/plugins/python/weechat-python.c:975
        rc = <optimized out>
        interpreter = 0x7ffff5bec910
        filename = 0x555555e69230 "/home/roald/.local/share/weechat/python/autoload/title.py"
#8  0x00007ffff66ea375 in weechat_python_unload_all ()
    at /builddir/weechat-4.1.1/src/plugins/python/weechat-python.c:1024
No locals.
[...]

PR #2222, opened 2024-12-29
python: set m_size for created modules to 0

trygveaa:
This value determines the size of the per-module memory area. Setting this value to -1 as it was before this change means that the module has global state and therefore does not support subinterpreters.

However, subinterpreters are used to run the Python scripts, so the weechat module has to support subinterpreters. Therefore we should set this value to 0 as no per-module memory is required.

This seems to fix the crash reported in #2046 without the need for the workaround added in commit 85c7494dc (it does for me when testing with Python 3.12.0 at least).

This change came up as a suggestion in cpython's issue tracker where it was pointed out that using modules with m_size set to -1 is not supported in subinterpreters. See these two comments:

https://github.com/python/cpython/issues/116510#issuecomment-2377915771 https://github.com/python/cpython/issues/116510#issuecomment-2389485369

It's not completely clear to me what is required for a module to support subinterpreters and re-initialization (which is required for setting m_size to 0), but https://peps.pythondiscord.com/pep-0489/ says:

    A simple rule of thumb is: Do not define any static data, except
    built-in types with no mutable or user-settable class attributes.

The only static data we define is of type int and str, so I think it should be fine.


trygveaa:
Interestingly/strangely, while confirming that this fixes the crash in #2046 without the need for the workaround, I noticed that the crash doesn't happen after commit 701f64afc17402449bf43a108591b126a3bf26f8 when the Ruby plugin is loaded.

So to test this crash I'm testing this with commit 85c7494dc7e4e283a5a6a0c6a89425a91e2e0a3a and commit c325dedf05e04b5791c3b3f36bd68094b2673ed0 reverted (or by just checking out v4.3.3) and with Python 3.12.6, as the crash doesn't occur with those commits or with Python 3.12.7. And to test what the Ruby plugin affects, I'm also testing without this PR, as this PR also fixes the crash.

Then if I don't load the Ruby plugin it crashes. And if I compile the Ruby plugin without commit 701f64afc17402449bf43a108591b126a3bf26f8 and load the Ruby plugin it also crashes. But if commit 701f64afc17402449bf43a108591b126a3bf26f8 is included and the Ruby plugin is loaded, it doesn't crash. I have no idea why the Ruby plugin affects this crash in the Python plugin.

PR #2230, opened 2025-01-22
core: use <stdbool.h> instead of typedef in ncurses-fake.h

luk1337:
Fixes the following error when building in Fedora rawhide:
error: ‘bool’ cannot be defined via ‘typedef’.

Likely GCC 15 related.


luk1337:
spotted here: https://koji.fedoraproject.org/koji/taskinfo?taskID=128160720

flashcode:
Thanks!

PR #2234, opened 2025-02-18
irc: let /server list|listfull take -connected option to filter the list

quite:


flashcode:
Thanks for the PR!

I merged with a few changes to display properly when no servers are found.
I also updated translations and made the French one.

PR #2267, opened 2025-07-26
Fill out program information fields in COPYING

web-coder-of-rpi:
I noticed that in the COPYING file some important fields were not filled out, like the ```<name>``` and ```<year>``` ones.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -631,8 +631,8 @@ to attach them to the start of each source file to most effectively
 state the exclusion of warranty; and each file should have at least
 the "copyright" line and a pointer to where the full notice is found.
 
-    <one line to give the program's name and a brief idea of what it does.>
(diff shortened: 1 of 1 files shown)


web-coder-of-rpi:
I’m not to sure about to put in the ```<name>``` field. It could be
```
https://github.com/weechat
``` 
or the creator of WeeChat’s name.

trygveaa:
These are not intended to be filled out in the COPYING file, it's just showing you what to include in other files. In the other files it should be filled out (though with the actual author names, not the repo URL), and WeeChat already does so (with SPDX tags as @flashcode mentions).

See for instance in the emacs repo, they're not filled in the COPYING file there either: https://github.com/emacs-mirror/emacs/blob/master/COPYING

PR #2268, opened 2025-08-23
Uprev most weechat dependencies

evelikov:
At some point last year, I updated some of the weechat dependencies and then dropped the ball for the rest :sweat_smile: 

This PR does a more thorough job updating curl, gnutls, gcrypt, lua and enchant, removing the respective dead code and/or workarounds.


flashcode:
Merged, thanks!

Issue #886, opened 2017-01-09
Add an option to store logs in UTC


nyuszika7h:
It would be nice to have an option to store logs in UTC, while still displaying time in the client according to the `$TZ` environment variable.

Also, perhaps add `%z` to the default `logger.file.time_format` to make the timezone obvious in logs, including DST transitions.

boreq:
>this function auto-detects the format (...) then we fallback to strptime with the format you configured

Given a non-zero amount of "magic" here I could see this leading to bugs in some cases but I think this shouldn't matter in reality.

>the date format for logs could be something more standard in logs

Yes, overall I'd support just using ISO 8601 in the logs, they are more likely to be read by a machine than a human I feel - and if a human encounters ISO 8601 they can very easily read it as well.

boreq:
I'd agree that the proposed format with UTC has the benefit of being able to use simple Unix tools for processing logs from different systems, otherwise you need more complex processing that actually compares the timestamps across timezones.

The problems I described with using UTC by default don't seem important in most cases. And any timezone info would be much better than the current default situation.

flashcode:
I just pushed a branch for tests: https://github.com/weechat/weechat/tree/logger-timezone

If possible, please test it, the new format is UTC, with a space between date and time.

It should be compatible with existing logs (if using the default format or similar one), tests with different timezones and formats are welcome.

PR #2269, opened 2025-08-30
Change default logger.file.time_format to %Y-%m-%d %H:%M:%S %z

boreq:
Currently the logs don't include the timezone information. This means that if the machine's timezone changes we won't know when the old messages were logged, if the logs are moved to a different computer we have the same problem, if we travel with our laptop we have the same problem, if we want to analyse them with other programs we can't easily read the timestamp either etc.

I believe this change isn't backward compatible but I have no other ideas other than just ripping this bandaid off because as it is I keep encountering this problem.

Related to #886.

### DIFF
--- a/src/plugins/logger/logger-config.c
+++ b/src/plugins/logger/logger-config.c
@@ -718,7 +718,7 @@ logger_config_init (void)
             N_("timestamp used in log files (see man strftime for date/time "
                "specifiers, extra specifiers are supported, see function "
                "util_strftimeval in Plugin API reference)"),
-            NULL, 0, 0, "%Y-%m-%d %H:%M:%S", NULL, 0,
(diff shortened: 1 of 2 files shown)


flashcode:
Hi,

Thanks for the PR, indeed adding timezone information makes sense in log files.

I made a proposal for preliminary changes before changing this format in the original issue: #886, please go on with comments in this issue.

I keep this PR open in the meanwhile.

boreq:
Yes: for context this is my suggestion that I submitted just because "maybe the solution really is that simple". I don't really know the codebase or the language so I have no idea what actually makes sense here.

flashcode:
Thanks for the PR, I close it as it's fixed via #886, and this specific commit: 21a958423ee1474bb7b02b38f87f2fd6d13af0c8.

The default format is now UTC: `%@%F %T.%fZ` (eg: `2025-09-02 21:11:39.816787Z`).

PR #2287, opened 2025-11-24
core/doc: update Serbian translation

eevan78:
Here is an update of the Serbian translation of documentation and core messages.


flashcode:
Thanks!

PR #2288, opened 2025-11-28
CMAKE: Support specifying a Ruby version

beholders-eye:
Hi, I'm a contributor at Exherbo GNU/Linux and a long time WeeChat user. In may I wrote this patch to fix weechat build in Exherbo, it basically enables you to pass RUBY_VERSION to CMake so it's explicity built against it, instead of using the system default one.


beholders-eye:
This is not needed, sorry to bother

PR #2289, opened 2025-11-30
api: fix buffer overflow when parsing date and time

luk1337:
sizeof(str_date2) > strlen (datetime) + 16 + 1.

### DIFF
--- a/src/core/core-util.c
+++ b/src/core/core-util.c
@@ -417,11 +417,12 @@ int
 util_parse_time (const char *datetime, struct timeval *tv)
 {
     char *string, *pos, *pos2, *pos_colon, *pos_hyphen, *pos_dot;
-    char str_usec[16], *error, str_date[128], str_date2[256];
+    char str_usec[16], *error, str_date[128];
     struct tm tm_date, tm_date_gm, tm_date_local, *local_time;
     time_t time_now, time_gm, time_local;
     long long value;
     int rc, length, use_local_time, timezone_offset, offset_factor, hour, min;
+    size_t string_size;
 
     if (!datetime || !datetime[0] || !tv)
         return 0;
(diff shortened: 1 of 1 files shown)


luk1337:
this was caught by one of hardening cflags fedora uses when building rpms:

```
export CFLAGS='-O2 -flto=auto -ffat-lto-objects -fexceptions -g -grecord-gcc-switches -pipe -Wall -Werror=format-security -Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3 -Wp,-D_GLIBCXX_ASSERTIONS -specs=/usr/lib/rpm/redhat/redhat-hardened-cc1 -fstack-protector-strong -specs=/usr/lib/rpm/redhat/redhat-annobin-cc1  -m64 -march=x86-64 -mtune=generic -fasynchronous-unwind-tables -fstack-clash-protection -fcf-protection -mtls-dialect=gnu2 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer  '
```

src/core/core-util.c
```diff
@@ -441,21 +442,23 @@ util_parse_time (const char *datetime, struct timeval *tv)
     if (pos_colon && !pos_hyphen)
     {
         /* add current date: "19:04:55" -> "2025-08-30T19:04:55" */
-        string = malloc (strlen (datetime) + 16 + 1);
+        string_size = strlen (datetime) + 16 + 1;
```
flashcode: I think the size here is still not good, it should be `strlen(str_date) + strlen(datetime) + 1`, according to what we put inside this string.

I'll take your changes and fix this.

luk1337: and bottom one should be `+ 9 + 1` instead of `+ 16 + 1`?

flashcode: Ah I was wrong, the str_date is less than 16 bytes in all cases.

So I think the initial code was not even buggy, but I understand the compiler may raise a warning due to `sizeof(str_date2)` being larger than the string.

luk1337: > `strlen(str_date) + strlen(datetime) + 1` makes tests fail here, not exactly sure why.

ah that's because i'm checking strlen too early.

flashcode:
I pushed a fix, please let me know if compiler is happy now.

I'll check which flags are used in your build, and add them if I can, so I can detect this as soon as I push code like this.

luk1337:
> I think another release is not needed, as the buffer overflow doesn't happen in practice. But let me know if you really need a fixed version (4.8.1).

I can just pull patch file locally, but would like if this landed into 4.8.1 whenever it's out.

flashcode:
> > I pushed a fix, please let me know if compiler is happy now.
> > I'll check which flags are used in your build, and add them if I can, so I can detect this as soon as I push code like this.
> 
> tests pass now, btw the flag is probably `-Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3`

It seems it's not this option, I'm checking.

flashcode:
Can I see the Redhat built output somewhere, with the compiler error?
Or can you paste it?
Thanks.

flashcode:
Ah I managed to reproduce, it happens in tests when we use `-O2` together with `-Wp,-U_FORTIFY_SOURCE,-D_FORTIFY_SOURCE=3`.

PR #2298, opened 2026-01-22
Build fix for Haiku

Begasus:


Begasus:
There is something around, haven't done it myself, but I'll poke around and get back to you on that.

PR #2317, opened 2026-05-08
relay: Change unix socket permisisons to 0770

WhatAmISupposedToPutHere:
Making the socket only accessible to the user complicates running headless weechat under a dedicated user, as it requires running nginx under the weechat user. 0770 permissions make it possible to add the nginx user to the weechat group and run both programs under separate users.


flashcode:
Hi,

Thanks for the PR!

I introduced a new option to configure the permissions on the Unix socket, `700` being the default value, and you can change the value to give access to the group, like this:

```
/set relay.network.unix_socket_permissions 770
```

PR #2320, opened 2026-06-01
core: fix possible integer truncation in function eval_string_split

aizu-m:
Same truncation as 0c29e5a, in the sibling split function: max_items from the max_items=N flag is parsed with util_parse_long, then passed to string_split() whose num_items_max is int. A value above INT_MAX (try ${split:1,,max_items=2147483648,a,b,c}) wraps to a negative int and becomes count_items inside string_split. Parse it with util_parse_int instead.


flashcode:
Thanks, I missed this one!

PR #2321, opened 2026-06-02
xfer: strip directory separator from remote nick in download filename

aizu-m:
A received DCC file's local path is built from the remote nick when xfer.file.use_nick_in_filename is enabled (default on). The filename is reduced to a basename in xfer_add, but the nick comes straight from the IRC message prefix with no validation, so a malicious server can send a DCC SEND whose sender nick contains a directory separator and have the file written outside xfer.file.download_path.


flashcode:
Thanks, merged!

PR #2322, opened 2026-06-04
Fix out-of-bounds read in irc_ctcp_recv_dcc quoted filename parsing

aizu-m:
Reading the DCC parser I noticed that when a quoted SEND/RESUME/ACCEPT filename ends the message, pos += 2 skips past the terminating NUL and the address scan then reads past the buffer, so reject a closing quote with nothing after it.


flashcode:
Thanks, merged!

PR #2325, opened 2026-06-08
core: fix buffer overflow in function network_pass_socks5proxy

aizu-m:
bound the configured proxy username and password before they are copied into the fixed stack buffer in network_pass_socks5proxy, otherwise a login longer than the buffer (a long password or token) overruns it while building the SOCKS5 auth request.


flashcode:
Thanks, merged!

PR #2328, opened 2026-06-20
doc/api: note that colors param is supported nicks endpoint

mhoran:
The `/api/buffers/{buffer_id}/nicks` endpoint supports the `colors` param but this was not noted in the docs.


flashcode:
Thanks, this parameter is indeed missing.

But `colors` is a query parameter, so I'll fix that and merge.

flashcode:
Merged!

PR #2331, opened 2026-06-28
relay/api: only decompress compressed messages

mhoran:
With permessage-deflate, RSV1 of the first fragment indicates whether or not the message is compressed [1]. If RSV1 is not set then the message should not be decompressed.

[1] https://datatracker.ietf.org/doc/html/rfc7692#section-6

### DIFF
--- a/src/plugins/relay/relay-websocket.c
+++ b/src/plugins/relay/relay-websocket.c
@@ -650,7 +650,7 @@ relay_websocket_decode_frame (const unsigned char *buffer,
     size_t size_decompressed;
     char *payload_decompressed;
     struct t_relay_websocket_frame *frames2, *ptr_frame;
-    int size, masked_frame, mask[4];
+    int size, compressed, masked_frame, mask[4];
 
     if (!buffer || !frames || !num_frames)
         return 0;
@@ -670,6 +670,10 @@ relay_websocket_decode_frame (const unsigned char *buffer,
             goto missing_data;
 
         opcode = buffer[index_buffer] & 15;
+
+	/* RSV1 of the first fragment indicates whether this message is
+	 * compressed */
+        compressed = (buffer[index_buffer] & 64) ? 1 : 0;
 
         /* check if frame is masked */
         masked_frame = (buffer[index_buffer + 1] & 128) ? 1 : 0;
@@ -777,9 +781,9 @@ relay_websocket_decode_frame (const unsigned char *buffer,
 
         /*
          * decompress data if frame is not empty and if "permessage-deflate"
-         * is enabled
+         * is enabled and the message is compressed
          */
-        if ((length_frame > 0) && ws_deflate && ws_deflate->enabled)
+        if ((length_frame > 0) && ws_deflate && ws_deflate->enabled && compressed)
         {
             if (!ws_deflate->strm_inflate)
             {


src/plugins/relay/relay-websocket.c
```diff
@@ -670,6 +670,7 @@ relay_websocket_decode_frame (const unsigned char *buffer,
             goto missing_data;
 
         opcode = buffer[index_buffer] & 15;
+        compressed = (buffer[index_buffer] & 64) ? 1 : 0;
```
mhoran, line 676: The spec says to look at `RSV1` on the first fragment of the message, so technically this is incorrect but I didn't want to move outside the `while` loop. I could just look at `buffer[0]` instead. Let me know what you'd prefer.

flashcode:
Hi @mhoran,

Thanks for the fix.

I'm surprised I didn't catch this when adding the API relay with permessage-deflate support, how did you notice that? A client sending non compressed frame causes problems in relay?

mhoran:
> I'm surprised I didn't catch this when adding the API relay with permessage-deflate support, how did you notice that? A client sending non compressed frame causes problems in relay?

Yes, when I tested weechatRN on Android (see https://github.com/mhoran/weechatRN/issues/118#issuecomment-4823834831) weechat rejected the initial message sent from the client. On Android, React Native uses [OkHttp](https://square.github.io/okhttp/) for WebSockets, which implements `permessage-deflate` but only compresses messages over [1 KiB](https://github.com/square/okhttp/blob/0cadfa2997513d20bf88ca530c963a1266f17af9/okhttp/src/commonJvmAndroid/kotlin/okhttp3/internal/ws/RealWebSocket.kt#L700). On iOS, React Native uses [SocketRocket](https://github.com/facebookincubator/SocketRocket), which does not implement `permessage-deflate`.

src/plugins/relay/relay-websocket.c
```diff
@@ -671,6 +671,10 @@ relay_websocket_decode_frame (const unsigned char *buffer,
 
         opcode = buffer[index_buffer] & 15;
 
+	/* RSV1 of the first frame indicates whether all frames of this message
+	 * are compressed */
+        compressed = (buffer[0] & 64) ? 1 : 0;
```
flashcode: @mhoran: I think this should be `bufferr[index_buffer]` here, because the `buffer` may contain different unrelated frames, and each frame may be or not compressed. In the other change below, the frame at `index_buffer` is decompressed if needed.

I think it works by chance because most of times you have only a single frame here, or all frames received have all the same bit RSV1 set.

mhoran: That is what I had tried originally but when I re-read the [spec](https://datatracker.ietf.org/doc/html/rfc7692#section-6) I believe this is correct:
>  A message with the "Per-Message Compressed" bit set on the **first fragment** of the message is called a "compressed message". [...] An endpoint receiving a compressed message decompresses the concatenation of the compressed data of the frames of the message by following the decompression procedure specified by the PMCE in use.

and under section [6.1](https://datatracker.ietf.org/doc/html/rfc7692#section-6.1):
> An endpoint MUST NOT set the "Per-Message Compressed" bit of control frames and non-first fragments of a data message.  An endpoint receiving such a frame MUST _Fail the WebSocket Connection_.

After reading the code, my understanding is that `buffer` contains frames (fragments, above) of the same message, not unrelated frames. Is my understanding incorrect?

FWIW both seem to work, but this seems to be an edge case in the spec and so looking at the first frame seems like it will be less error prone.

flashcode: The buffer in input of the function is just an array of bytes read from a remote machine, and the function returns a list of frames. So each frame may be or not compressed, and RSV1 is specific to each frame.

So for me the RSV1 flag should be read by `buffer[index_buffer]`.

mhoran: I see, WeeChat doesn't support [fragmentation](https://datatracker.ietf.org/doc/html/rfc6455#section-5.4), and expects each frame in the buffer is expected to be a full message. In that case I agree with `buffer[index_buffer]`.

I haven't run into any issues with fragmentation not being implemented, but did notice that the `FIN` bit was not being checked, so I was wondering about that. For now I will make the code change and leave a note that this needs to be fixed if fragmentation support is added.

flashcode:
Thanks, merged!

I updated the comment and removed some trailing tabs.

PR #2335, opened 2026-07-15
core: fix integer overflow in eval of ${hide} and ${base_encode}

arib06:
guard the output size in eval_string_hide and eval_string_base_encode: the character count times the hide/encode factor is computed in int and can wrap, so malloc returns a tiny buffer that the following writes overrun.


flashcode:
Thanks, merged with some minor style changes in the unit test added.

PR #2338, opened 2026-07-16
perf: Improve Weechat upgrade processing

Jonher937:
## Background

After experiencing a long /upgrade save+load I started looking at the steps taken during the restore of the upgrade files and what changes could be made to speed it up.
For small buffers and scrollback it won't matter much, but the bigger the scrollback, the more that read/write pass costs. This branch goes after overhead from a few different angles.

Upgrade files were being read and written as a huge number of tiny fread/fwrite calls using the default stdio buffer size, so I bumped that up to 256 KB to cut down on the actual syscalls happening underneath. I also noticed every single field read was calling ftell() just to record a position that's only ever used if something goes wrong and an error needs to be reported, so instead of asking the OS for the current position every time, it now just keeps a running counter itself.

Then there was a memory/allocation problem: every field coming out of the upgrade file was effectively being copied twice, once when it was read off disk into a temporary buffer, and again when that buffer got handed to the infolist via strdup(). Since the first copy was just going to be thrown away, I added "take ownership" versions of the infolist constructors so the data (and its name) gets handed over directly instead of duplicated a second time.

On the GUI side restoring lines was redoing work that had already been done once and saved. gui_line_new() was recomputing the highlight state and the colorized timestamp string for every restored line, even though both of those were already sitting in the upgrade file and were about to be overwritten anyway.
I added a way to pass those in as "already known" so that recomputation gets skipped. I also found that appending lines during restore was scanning from the very start of the line list every single time to find where to insert, which is fine for one line but turns into quadratic behavior across an entire restored buffer, so there's now a fast path for the common case where lines come in already in order.
And restoring lines was also triggering hotlist updates that were guaranteed to be wasted, since the hotlist gets rebuilt from its own saved state right after anyway, so that got skipped too during replay.

## A bug found along the way

While working on the highlight-state change, I realized that when a line got restored, WeeChat would compute its colorized timestamp based on the highlight state at that moment, but then immediately overwrite the highlight state with the value that had actually been saved, without ever going back and recomputing the timestamp's colors.
So a restored line's timestamp could end up colorized as if it weren't a highlight when it actually was, or the other way around. Since I was already threading the "known" highlight and timestamp values through this code for performance reasons, the fix fell out naturally: pass the saved values straight through instead of computing them, which is both faster and no longer produces a timestamp colorized for a highlight state that's about to be replaced.

### DIFF
--- a/src/core/core-infolist.c
+++ b/src/core/core-infolist.c
@@ -277,7 +277,7 @@ infolist_new_var_buffer (struct t_infolist_item *item,
 }
 
 /*
- * Creates a new string variable in an item, taking ownership of "value"
+ * Creates a new integer variable in an item, taking ownership of "name"
  * instead of copying it (the infolist will free it later).
  *
  * INTERNAL USE ONLY, see comment in core-infolist.h.
@@ -286,8 +286,8 @@ infolist_new_var_buffer (struct t_infolist_item *item,
  */
 
 struct t_infolist_var *
-infolist_new_var_string_take_ownership (struct t_infolist_item *item,
-                                        const char *name, char *value)
+infolist_new_var_integer_take_name_ownership (struct t_infolist_item *item,
+                                              char *name, int value)
 {
     struct t_infolist_var *new_var;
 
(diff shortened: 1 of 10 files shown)


flashcode:
Merged with changes on a temporary branch "perf-upgrade" for tests.

Now I'm merging #2339 that contains the additional optimizations.

flashcode:
Branch perf-upgrade has been merger into main.

PR #2339, opened 2026-07-16
perf: implement infolist cursor for upgrade files

Jonher937:
Restoring session state from a WeeChat upgrade file leans heavily on infolists to read back saved objects field by field.
The problem is that every single field lookup walked the item's variable list from the very beginning, so an object with a dozen or so fields paid a quadratic cost just to read itself back.
On buffers with a lot of scrollback, that adds up.

This change adds a small cursor to each infolist item that remembers where the last lookup landed, so the next lookup can pick up from there instead of starting over.
Since most callers already read fields roughly in the order they were written, this turns the common case into a fast forward scan, while still falling back to a full search if something is read out of order.

To get the full benefit, the buffer-line restoration code was also tidied up: it now reads all of a formatted line's fields into local variables first, in the exact order they were originally saved, rather than pulling them inline as function arguments (where C doesn't guarantee the order they'd actually be evaluated in). That makes the lookups line up with the cursor's forward-scanning assumption almost every time, and as a side benefit makes the read order deterministic regardless of compiler.

Net effect: faster upgrades/reloads for users with large scrollback history.

Depends-on: #2338

### DIFF
--- a/src/core/core-infolist.c
+++ b/src/core/core-infolist.c
@@ -277,7 +277,7 @@ infolist_new_var_buffer (struct t_infolist_item *item,
 }
 
 /*
- * Creates a new string variable in an item, taking ownership of "value"
+ * Creates a new integer variable in an item, taking ownership of "name"
  * instead of copying it (the infolist will free it later).
  *
  * INTERNAL USE ONLY, see comment in core-infolist.h.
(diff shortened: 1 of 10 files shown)


flashcode:
Merged on the temporary branch perf-upgrade for tests.

flashcode:
Branch perf-upgrade has been merger into main.

PR #2340, opened 2026-07-17
logger: fix path traversal in log file name built with mask

arib06:
expand the buffer local variables in each directory level of the mask separately in logger_get_mask_expanded, otherwise a \001 byte coming from a local variable (a channel name is chosen by the server) is missed by the separator replacement and the restore step turns it into a real directory separator, writing the log outside logger.file.path.


flashcode:
Hello,

Thanks for the fix, I'll check this as soon as possible.

Please note that all security fixes must be submitted here (not as standard issue): https://github.com/weechat/weechat/security (see also the [WeeChat security page]( https://weechat.org/doc/weechat/security/)).

So this can be made public only once fixed and released, to prevent any exploit on running WeeChat instances.

flashcode:
Thanks, merged with some minor changes in comment.
I also added a new commit with non-regression unit test.
And I backported the fix on branch 4.9.

PR #2341, opened 2026-07-22
core: grow/shrink shared hashtable buckets based on load factor

Jonher937:
Add hashtable_rehash() to core-hashtable.c: reallocates the bucket array to a new size and reinserts every item at its sorted position in its new bucket, preserving the invariant hashtable_get_item's early-exit comparison relies on.

hashtable_set_with_size() calls it once items_count exceeds twice the current size (doubling), so hashtable_new()'s size argument is now just an initial size rather than a fixed ceiling. hashtable_remove() calls it symmetrically once items_count drops below a quarter of the current size (halving, down to a floor of 8 buckets), so a hashtable that grew for a transient spike does not keep the larger array indefinitely. The 2x/4x gap between the grow and shrink thresholds is a deliberate hysteresis margin: it keeps a single add/remove pair near either boundary from flipping the table back and forth between growing and shrinking.

Avoiding long strcmp-based collision chains for any workload with many unique shared strings (eg. large upgrade files with many distinct tag combinations) without over-allocating for smaller ones, in either direction as the hashtable's contents change over its lifetime.

### DIFF
--- a/src/core/core-hashtable.c
+++ b/src/core/core-hashtable.c
@@ -40,6 +40,30 @@
 #include "../plugins/plugin.h"
 
 
+/*
+ * Grow (double) the internal array when the average chain length
+ * (items_count / size) exceeds this value.
+ */
+#define HASHTABLE_LOAD_FACTOR_MAX 2
+
+/*
+ * Shrink (halve) the internal array when the average chain length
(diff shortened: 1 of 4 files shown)


Jonher937:
Before this changed it took ~650seconds to restore a large upgrade file, with this change we're down to ~30seconds.

PR #2342, opened 2026-07-23
core: fix buffer overflow in display of time in chat area

arib06:
size text_time2 for up to 4 bytes per char (a color code plus the char) in gui_chat_get_time_string, otherwise a buffer_time_format that alternates digits and non-digits overflows the fixed stack buffer while coloring the time.


flashcode:
Thanks, merged on branches main, 4.9 and 4.6 with one minor change: in the unit test, the value of the changed option is restored with a call to `config_file_option_reset()` like it's done in other existing tests (so no need to save the old value and restore it).

Issue #2170, opened 2024-08-13
api relay: implement a "read marker" so notifications do not get replayed on every reconnect


NetSysFire:
### Feature description

Notifications are entirely handled on the clientside by the connecting weechat. When you reconnect, whether that be because of a network issue, an /upgrade or something else, all notifications will be replayed, leading to a flood of notifications you have to manually dismiss, especially if the remote buffer size is big.

Pinged 5 days ago in a quiet channel? You will be pinged again on a reconnect. Logged in on e.g libera.chat? The "Last login from" message from SaslServ will always be replayed.

The solution for that, which is apparently already implemented for the irc relay, is probably a read marker system, which is purely on the client side which tracks the seen status of buffers. weechat already correctly dismisses notifications if you switch to that buffer, so maybe it can be built upon that. It should track the seen status per buffer as a timestamp and when a reconnect happens to the same remote, notifications older than the timestamp will not be replayed. Maybe it should also save that timestamp so it works even when you /quit the connecting instance.

ccxcz:
For seamless experience storing read markers on host (API server) Weechat instance and notifying clients of read status changes is necessary for almost every non-trivial use-case.
Switching between multiple clients, having clients that can't persist their state, and using clients that aren't online all the time will all require server-managed hotlist/read status.

Less seamless solution but much simpler one would be to implement ability to copy remote hotlist to client.
This could be exposed as command and ideally also done on connect to a remote.
That would eliminate the most annoying part of the problem which is highlights and unread messages reappearing every reconnect.
I'd be comfortable enough with marking buffers read manually via command to the remote, at least as interim solution.
Also I'd appreciate option to mark a buffer as read each time a message from my nickname appears on it — whether that's another client on same bouncer, log replay, or something else. That should also cut down the hotlist noise quite a bit by itself.

RyanSquared:
I think I have a rough plan on how this can be implemented.

## Sync: Remote to Client

In the `weechat` protocol, infolist_buffer has an entry first_line_not_read. This value can be added to `GET /buffer`, which can be used in association with the `relay_remote_line_id_` tag in lines. If the line ID is below the first line not read, do not add it to the hotlist. Then, locally, once we have a `relay_remote_line_id_` that is higher than the `first_line_not_read`, lines can be added to the hotlist again. This would likely need to be stored as a separate item in the infolist, since remote line IDs and local line IDs may differ.

## Sync: Client to Remote

[...]

PR #2343, opened 2026-08-04
relay: skip adding notify tag when already read on remote

RyanSquared:


RyanSquared:
One caveat identified: If I mark a channel read on server, then something happens that is notify_none on the server, there's no indication to the client that it also should be notify_none on the client. For example, I disconnected from a server. I do not believe it added new entries to the hotlist on the server, but when connecting with the client, I did see it get put in the hotlist. Given it's still a low notify, this is still a significant upgrade over the current setup.

RyanSquared:
I've added another commit to address this, which unconditionally adds `notify_none` to any previously read line.

RyanSquared:
Closes: https://github.com/weechat/weechat/issues/2170

RyanSquared:
On the server:

```
/alias add markread /buffer set unread;/buffer set hotlist -1
```

On the client:

```
/trigger addreplace sync_remote_read signal "buffer_switch" "${buffer.full_name} =~ ^relay.remote.${relay_remote}" "" "/remote send ${relay_remote} {\"request\": \"POST /api/input\", \"body\": {\"buffer_name\": \"${relay_remote_name}\",\"command\": \"/markread\"}}" "ok" "none"
```

RyanSquared:
Now, the only missing feature is updating the read marker on clients when it's marked read on the server. I think that's a bit more complicated, so I may revisit that at a later date.

src/plugins/relay/api/remote/relay-remote-event.c
```diff
@@ -929,6 +929,8 @@ RELAY_REMOTE_EVENT_CALLBACK(buffer)
     snprintf (str_number, sizeof (str_number), "%lld", id);
     weechat_hashtable_set (buffer_props,
                            "localvar_set_relay_remote_id", str_number);
+    weechat_hashtable_set (buffer_props,
+                           "localvar_set_relay_remote_name", name);
```
flashcode, line 933: What is the purpose of this new buffer local variable? It seems not used.

Is it mandatory for this PR?

flashcode (changes requested):

flashcode:
The other changes seem OK to me, I'll make just a few changes and test:

- remove the "static" function (not used elsewhere in plugins, so for consistency)
- fix bug with "notify_" tag, a comma may be added alone (due to the "continue"), without tag after
- fix long lines
- add unit tests.

src/plugins/relay/api/remote/relay-remote-event.c
RyanSquared: > What is the purpose of this new buffer local variable? It seems not used.

It is critical for the trigger I've mentioned, without which this PR is only half functional. If you have an alternative proposal for how to sync local read markers to the remote, I'll remove it (or you can drop the third commit, it's self-contained), but this PR feels only half complete without it.

flashcode: OK, then I will first merge this PR (without the last commit) and implement the second fix as soon as possible, with a clean solution in the relay plugin itself (so with no use of triggers).

src/plugins/relay/api/remote/relay-remote-event.c
```diff
@@ -64,6 +64,26 @@
...
 
 
+/*
+ * Check if a line in a buffer has been marked read on the remote.
+ */
+static int
+relay_remote_line_is_already_read (struct t_gui_buffer *ptr_buffer, int line_id)
+{
+    const char *ptr_val;
+    long long last_read;
```
flashcode, line 74: The "last_read_line_id" is an "int" and not a "long long" (the id in a line is an int). This is an issue, see the next comment below.

src/plugins/relay/api/remote/relay-remote-event.c
```diff
@@ -64,6 +64,26 @@
...
+
+    if (!ptr_buffer || line_id < 0)
+        return 0;
+
+    ptr_val = weechat_buffer_get_string (ptr_buffer, "localvar_relay_remote_last_read_line_id");
+    if (!ptr_val || !weechat_util_parse_longlong (ptr_val, 10, &last_read))
+        return 0;
+
+    return line_id <= last_read;
```
flashcode, line 83: This will not always work: the id of a line is an int and when INT_MAX is reached, the id restarts to 0.

There should not be two lines with the same id (because INT_MAX should always be higher than the number of lines in the buffer), but comparing two id doesn't always indicate which line is older in the buffer.

One solution would be to convert the line id to "long long", but this has two impacts: memory (each line will use more memory, and WeeChat stores a lot of lines) and all consumers of this data.

The other solution would be to check if a line is read with another test, I'll check how to make this.

flashcode:
I have pushed a working branch "relay-remote-lines-notify" with the contents of this PR and the additional fixes. I'll work on a full solution that includes fix on the "check if line is read" and the sync local read markers to the remote.

flashcode:
@RyanSquared: I pushed a branch "line-id-timestamp" with additional fixes, could you please try this branch and check everything is OK?

Two important notes:

- the commit for the trigger is not present there, I'll check how to implement a clean solution
- if you `/upgrade` to this branch, you can not `/upgrade` again to the main branch, because the lines have an incompatible format, so better try with a new and separate WeeChat config

flashcode:
The branch has been merged into main, and then removed.

PR #2347, opened 2026-08-16
relay: bound partial client messages

acts-1631:
An unauthenticated client that can reach a configured relay listener can send data without a line terminator. The text relay path retained the incomplete message by reallocating and appending on every read, while the client remained in AUTHENTICATING until the authentication timeout. This allowed memory usage to grow without a per-message bound and could exhaust WeeChat before authentication completed.

This change bounds retained partial text at 8 MiB before reallocating. Extra data is ignored once the limit is reached, matching the existing protection in the HTTP relay path.


flashcode:
Thanks, merged!

PR #2348, opened 2026-08-16
xfer: keep received filenames within download path

acts-1631:
An incoming DCC filename is eventually used to build the local destination path. The existing receive path strips only the platform-native directory separator, so an alternate separator can remain in the filename and be interpreted as a path separator by file APIs on platforms that accept both forms. If file transfers are automatically accepted or the sender is configured as an auto-accepted nick, this could place received data outside the configured download directory.

This change keeps only the final filename component before constructing the local path and strips both slash forms. The configured download directory, remote-nick prefix, and collision suffix behavior are preserved.


flashcode:
Thanks, merged!

PR #2349, opened 2026-08-20
xfer: bound unterminated DCC chat messages

acts-1631:
A DCC chat peer can exhaust WeeChat memory by sending an unterminated message.

When IRC receives a DCC CHAT request, `irc-ctcp.c` creates a `chat_recv` transfer. After the user accepts it, or when `xfer.file.auto_accept_chats` is enabled, `xfer-network.c` installs `xfer_chat_recv_cb` on the active socket. That callback concatenates each read without a newline into `xfer->unterminated_message`. Unlike the IRC and relay message accumulators, the DCC chat path has no length limit, and the request timeout no longer applies after activation.

This patch caps the accumulated message at 64 KiB and closes the transfer before the concatenation would exceed the cap. This bounds the retained state and the temporary allocation used while combining reads.


flashcode:
Thanks, fixed with some changes:

- discard partial message and keep DCC chat open when size limit is reached
- increase limit to 8MB (to match the behavior in relay plugin)
- add a unit test.

PR #2350, opened 2026-08-20
irc: reject invalid ECDSA SASL challenge data

acts-1631:
A malformed IRC SASL ECDSA challenge can terminate WeeChat.

Server-supplied `AUTHENTICATE` data reaches
`irc_sasl_mechanism_ecdsa_nist256p_challenge()`. Invalid base64 returned
`-1`, which was assigned to the unsigned `gnutls_datum_t.size` field
before signing. This could make GnuTLS process an invalid oversized datum
and crash the client.

Reject non-positive decode results before loading the key or signing the
challenge.


flashcode:
Thanks, merged, and a unit test has been added.

PR #2351, opened 2026-08-20
script: reject path components in repository names

acts-1631:
A crafted script repository entry can escape WeeChat’s script cache when the user installs it.

The repository parser copies the untrusted <name> field into name_with_extension. /script install then passes script_config_get_script_download_filename() as curl’s file_out path. Without filename validation, ../ or Windows separators could overwrite files outside script.scripts.path.

Reject names containing path or drive separators before adding them to the repository. Normal script names remain unchanged.


flashcode:
Thanks, merged with some minor changes:

- add a description of the returned value in the new function
- removed the "static" for the function
- simplified function: use `strpbrk()` instead of a loop
- added a unit test.

PR #2352, opened 2026-08-20
xfer: require expected state for DCC resume

acts-1631:
A DCC peer can cause WeeChat to accept an incoming file without the user accepting it.

DCC ACCEPT messages reach xfer_start_resume_cb(), which looks up transfers by server, type, and port. xfer_search() assigned XFER_STATUS_CONNECTING instead of comparing it, allowing a waiting transfer to start automatically and bypass manual acceptance.

Compare the status so resume callbacks only operate on transfers already in the expected state.


flashcode:
Thanks, merged, with a unit test added.

flashcode:
@acts-1631: thanks for the report of vulnerabilities. For any future security vulnerability reported, please open an advisory here: https://github.com/weechat/weechat/security

That way, it can be made public once the version is released.

PR #2355, opened 2026-08-28
irc: add option irc.look.buffer_switch_connect

marinelli:
### Add option to disable auto-switch to server buffer on connect

Connecting to a server (e.g. `/connect -all`) always switches the display to the new server buffer, with no way to opt out — unlike `/join`, which already has this toggle for channels.

Adds `irc.look.buffer_switch_connect` (default: `on`) to gate this behavior in `irc_server_connect()`. Set to `off` to stay on the current buffer when connecting.


flashcode:
Thanks, merged!

While testing this new option, I noticed a bug with the active buffer when a buffer is merged, so I fixed it in a separate commit.

PR #2356, opened 2026-09-03
relay: reject invalid WebSocket control frames

acts-1631:
The relay WebSocket decoder accepts control frames using the general 8 MiB payload limit instead of RFC 6455’s 125-byte limit. It also accepts fragmented or RSV1-marked control frames.

On the `weechat` and `irc` WebSocket endpoints, the handshake completes before client authentication. An unauthenticated remote client can send an oversized PING, which `relay_client_read_websocket_frames` echoes as a PONG. Because the socket is nonblocking and `relay_client_outqueue_add` has no bound, repeated requests can consume process memory when the client does not read responses.

Reject fragmented, compressed, or oversized control frames before payload allocation or inflation. Invalid frames then use the existing decoder error path and disconnect the client.


flashcode:
Thanks, merged with a unit test added.

PR #2357, opened 2026-09-03
relay: cap client output queue

acts-1631:
A relay WebSocket client can send valid PING control frames before authenticating. `relay_client_read_websocket_frames` responds to each PING through `relay_client_send`. If the client stops reading, the nonblocking socket eventually returns `EAGAIN`, and `relay_client_outqueue_add` appends each PONG to an unbounded queue.

An unauthenticated remote client on the `weechat` or `irc` endpoint can use this to grow WeeChat’s memory during the authentication window, causing denial of service. Unlike #2356, the frames used here comply with the WebSocket control-frame limits.

Track queued output bytes and disconnect clients that exceed the limit.


flashcode:
Thanks, merged with an additional fix and a unit test added.

nuom17568-hue (commented):
一帆风顺

PR #2358, opened 2026-09-04
irc: remove duplicate check on batch tag

darkdi:
The batch tag guard in `irc_protocol_tags_cmd` tests `ptr_tag_batch` twice. I went looking for whether the second operand was meant to be `ptr_tag_batch[0]`, the way `command` and `extra_tags` are tested a few lines further down, but the other place that reads the same tag, near the end of `irc_protocol_recv_command`, checks the pointer only, so I dropped the repeat rather than turning it into an emptiness check.


flashcode:
Thanks, merged!

